PR #1, opened 2021-02-10
Initializing repo

alexanderConstantinescu:
This commit initializes the repository 

It adds:
- OWNERS file
- Makefile
- Dockerfile
- .gitignore
- Base file for binary

/assign @abhat @danwinship @tssurya @squeed


Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.8 AS builder
```
danielmellado, line 1: I'd also add a rhel7 builder, which might be needed later on

i.e.

`FROM registry.ci.openshift.org/ocp/builder:rhel-7-golang-1.15-openshift-4.7 AS rhel7`

alexanderConstantinescu: I don't think it will. This component will only be deployed on OCP >= 4.8 which only runs rhel8.

OWNERS
```diff
@@ -0,0 +1,24 @@
...
+  - abhat
+  - danwinship
+  - dcbw
+  - knobunc
+  - trozet
+  - rcarrillocruz
+  - squeed
+
+# Bugzilla info
+component: Networking
```
danwinship, line 24: I think you need to specify subcomponent as well. Even though there isn't a correct answer. Pick either openshift-sdn or ovn-kubernetes

alexanderConstantinescu: Done

alexanderConstantinescu:
@openshift/openshift-team-sdn PTAL

tssurya:
/lgtm

Would be nice to have a brief README asap.

PR #4, opened 2021-02-16
CloudPrivateIPConfig CRD definition and project dependencies 

alexanderConstantinescu:
This PR adds the `CloudPrivateIPConfig` CRD definition and initializes the project repo, essentially importing all vendor files required for `controller-gen` to generate the necessary client-go building blocks. 

The commits that are interesting for review are: 

- `Adding CRD definition for cloud private IP config`
- `Adding codegen script for CRD definitions` - which is quite a blatant copy of https://github.com/openshift/ovn-kubernetes/blob/master/go-controller/hack/update-codegen.sh

I have written the controllers for this CRD, but I will post that in a later PR once this gets merged as to make the review process easier. 

/assign @squeed @abhat @tssurya @danwinship


alexanderConstantinescu:
```
ci/prow/unit — Job succeeded.
```

:1st_place_medal:

pkg/crd/cloudprivateipconfig/v1/types.go
```diff
@@ -0,0 +1,62 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Specification of the desired private IP request(s).
+	Spec CloudPrivateIPConfigSpec `json:"spec"`
+	// Observed status of CloudPrivateIPConfig. Read-only.
+	// +optional
+	Status CloudPrivateIPConfigStatus `json:"status,omitempty"`
+}
+
+type CloudPrivateIPConfigSpec struct {
```
abhat, line 33: the type doesn't have to say `Spec` could be just `CloudPrivateIPConfig`

pkg/crd/cloudprivateipconfig/v1/types.go
```diff
@@ -0,0 +1,62 @@
...
+	// The list of per node requested private IPs and their corresponding node assignment.
+	Items []CloudPrivateIPConfigItem `json:"items"`
+}
+
+type CloudPrivateIPConfigStatus struct {
+	// The list of per node status of private IP assignments
+	// executed. In case an assignment fails: the item will be omitted from the
+	// status.items. I.e: if all went well the `.spec` should
+	// equal `.status`
+	Items []CloudPrivateIPConfigItem `json:"items"`
```
abhat, line 43: would  the status be something more than the list of items as in the `Spec`? I would imagine a struct that has conditions and/status such as `Assignment Successful`, `Assignment requested`, `Assignment Pending` etc might perhaps make more sense?

pkg/crd/cloudprivateipconfig/v1/types.go
```diff
@@ -0,0 +1,62 @@
...
+// +genclient:noStatus
+// +resource:path=cloudprivateipconfig
+// +kubebuilder:resource:shortName=cpip,scope=Cluster
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +kubebuilder:printcolumn:name="Node Request",type=string,JSONPath=".spec.items[*].node"
+// +kubebuilder:printcolumn:name="IP Request",type=string,JSONPath=".spec.items[*].ip"
+// +kubebuilder:printcolumn:name="Node Assignment",type=string,JSONPath=".status.items[*].node"
+// +kubebuilder:printcolumn:name="IP Assignment",type=string,JSONPath=".status.items[*].ip"
+// CloudPrivateIPConfig is a CRD allowing the user to assign private
+// IP addresses to the primary NIC on cloud VMs. This is done by
```
abhat, line 18: nit: verbiage refinement.
Suggested verbiage: 
`CloudPrivateIPConfig is a CRD type that allows users to create a config request for assigning private IP addresses to a NIC associated with a cloud VM.` 
The part about how to do it can be in the type of the `Spec` field

pkg/crd/cloudprivateipconfig/v1/types.go
```diff
@@ -0,0 +1,62 @@
...
+	// The list of per node status of private IP assignments
+	// executed. In case an assignment fails: the item will be omitted from the
+	// status.items. I.e: if all went well the `.spec` should
+	// equal `.status`
+	Items []CloudPrivateIPConfigItem `json:"items"`
+}
+
+type CloudPrivateIPConfigItem struct {
+	// Node name
+	Node string `json:"node"`
```
abhat, line 48: If a VM has more than one NICs provisioned, from a cloud providers' perspective how will we choose what NIC to assign this private IP to?

pkg/crd/cloudprivateipconfig/v1/types.go
```diff
@@ -0,0 +1,62 @@
...
+	// IP address - can be IPv4 or IPv6
+	IP string `json:"ip"`
```
abhat, line 50: maybe add a kubebuilder validation for the IPs?

pkg/crd/cloudprivateipconfig/v1/types.go
tssurya: The CR is not user facing right ? would only allow ovnkube-master to create the CR right ?

pkg/crd/cloudprivateipconfig/v1/types.go
```diff
@@ -0,0 +1,62 @@
...
+	Spec CloudPrivateIPConfigSpec `json:"spec"`
+	// Observed status of CloudPrivateIPConfig. Read-only.
+	// +optional
+	Status CloudPrivateIPConfigStatus `json:"status,omitempty"`
+}
+
+type CloudPrivateIPConfigSpec struct {
+	// The list of per node requested private IPs and their corresponding node assignment.
+	Items []CloudPrivateIPConfigItem `json:"items"`
+}
```
tssurya, line 36: validate for `minItems: 1` so that this field is unset, in the yaml, we don't allow its creation. Although maybe since this is not user facing we don't have to worry about this?

pkg/crd/cloudprivateipconfig/v1/types.go
tssurya: actually yea, else we would just be duplicating what is there in the request item fields right? Having status fields would also align with what the field intends to do.
maybe:
```
status: [
  {
    assignment: <nodename>-<ip-address>
    status: success
  },
  {
    assignment: <nodename>-<ip-address>
    status: failure
  },
]
```

pkg/crd/cloudprivateipconfig/v1/types.go
tssurya: I tried to do this for the egr crd, and failed because its hard to accommodate ipv6 validater, although I'd recommend at least adding some basic validation for both nodename and ip to check its not empty strings.

tssurya (commented):
So if we take Azure as an example, on the enhancement you mentioned a couple of other fields like rg_name, nic and "unique name" that we expect the cloud network config controller to find out/create before...will that be done as a part of the controller's processing of CR?

pkg/crd/cloudprivateipconfig/v1/types.go
alexanderConstantinescu: It can't be `CloudPrivateIPConfig`, because that would conflict with the primary struct name.

pkg/crd/cloudprivateipconfig/v1/types.go
alexanderConstantinescu: I am not sure it's needed, in the enhancement proposal I wrote: 

```
In case an assignment is unsuccessful: the `.status.assignment[?]` corresponding
to that `.spec.requests[?]` will not be set. This is done as to explicitly
notify the network plugin of which (if not all) requests have failed by still
issuing an UPDATE on the object.
```

I don't really see the point of a text valued status, since the network plugin will not proceed to finalizing its setup until the CNCC finishes its, and the network plugins don't have a text valued status for their egress IP assignment.
```
+--------------+                                   +-------------+                         +--------------+
|              |                                   |             |                         |              |
|              |                                   |             |                         |              |
|              |         CPIPC - Request           |             |                         |              |
|              |   +-----------+------------>      |             |                         |              |
|              |                                   |             |                         |              |
|              |                                   |             |        Cloud call       |              |
|              |                                   |             |  +--------------------> |              |
|    Network   |                                   |     CNCC    |                         |              |
|    Plugin    |                                   |             |                         |      Cloud   |
|              |                                   |             |                         |              |
|              |                                   |             |                         |              |
|              |                                   |             | Wait for cloud response                |
|              |                                   |             |   <-------------------+ |              |
|              |                                   |             |                         |              |
|              |                                   |             |                         |              |
|              |       CPIPC - status update       |             |                         |              |
|              |                                   |             |                         |              |
+--------------+    <-----------------------+      +-------------+                         +--------------+
```
I am fine with the network plugin just notifying the user that the CNCC couldn't finalize the setup.

pkg/crd/cloudprivateipconfig/v1/types.go
alexanderConstantinescu: > The CR is not user facing right ?

Yes

> would only allow ovnkube-master to create the CR right ?

We could make it so that only the service accounts associated with the network plugins are allowed to create the CR.  

I however agree with @abhat that this sentence could benefit from some re-writing.

pkg/crd/cloudprivateipconfig/v1/types.go
alexanderConstantinescu: We only perform the assignment to the primary NIC on the VM (index: 0). It might not be explicitly clear from the enhancement proposal, but it is "somewhat" stated in the non-goals: 

```
Non-Goals

...
* Assignment of additional NICs to cloud VMs
...
```

pkg/crd/cloudprivateipconfig/v1/types.go
alexanderConstantinescu: Yes, IPv6 is the main reason I chose not to do so. The validation of all fields is/will be done in the controller.

alexanderConstantinescu:
> So if we take Azure as an example, on the enhancement you mentioned a couple of other fields like rg_name, nic and "unique name" that we expect the cloud network config controller to find out/create before...will that be done as a part of the controller's processing of CR?

Yes. The CNCC will find out all of that out based on the `v1.Node` object retrieved from the node name passed through the spec. `v1.Node` contains a `.spec.ProviderID` field which has all the relevant information needed to retrieve the VM from the cloud API.

pkg/crd/cloudprivateipconfig/v1/types.go
alexanderConstantinescu: Sure, this seems doable.

alexanderConstantinescu: Added `minItems: 1` and `uniqueItems: true`

pkg/crd/cloudprivateipconfig/v1/register.go
```diff
@@ -0,0 +1,29 @@
+package v1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var (
+	GroupName          = "network.openshift.io"
```
squeed, line 10: This group is already in use - you'll need to pick a new one.

pkg/crd/cloudprivateipconfig/v1/types.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +genclient
+// +genclient:nonNamespaced
+// +genclient:noStatus
+// +resource:path=cloudprivateipconfig
+// +kubebuilder:resource:shortName=cpip,scope=Cluster
```
squeed, line 11: Don't use a short-name; this isn't a user-facing type.

pkg/crd/cloudprivateipconfig/v1/register.go
alexanderConstantinescu: ack will change

pkg/crd/cloudprivateipconfig/v1/types.go
alexanderConstantinescu: I meant to have it like this a short-cut for us when debugging this stuff. What does it hurt if it's specified?

squeed: shortnames are a limited space. In the past, we've had pushback when adding them, unless they are a core type that a user would type themselves.

For debugging, just set a shell alias.

pkg/crd/cloudprivateipconfig/v1/types.go
abhat: this is mostly for debuggability as an end-user. imagine a flow where the CNCC tried to do its thing and failed. Of course the network plugin won't proceed, but what is the indication to the end-user as to what went wrong?

pkg/crd/cloudprivateipconfig/v1/types.go
abhat: there are some places I think in openshift/api/network/v1 where we may have some precedent for v6 validator. If it's too much hassle, we can leave it. but i liken the kubebuilder validations to compile time checks vs. run time. Better to fail fast.

abhat: also, does it make sense to split this into a v4 field and a v6 field to make things clear?

pkg/crd/cloudprivateipconfig/v1/types.go
alexanderConstantinescu: Well, if we've decided that this component is not user-facing, then there wouldn't be much more information needed for the user to debug. All they care about is if egress IP works or not, and if it doesn't: create a ticket to us and we would know to look in the CNCC.

pkg/crd/cloudprivateipconfig/v1/types.go
alexanderConstantinescu: I've had a look and didn't find any IPv6 validators, maybe because not much IPv6 work has been done for openshift-sdn? Do you know of one such place? 

> also, does it make sense to split this into a v4 field and a v6 field to make things clear?

Nothing else in the Kubernetes eco-sphere makes that distinction, I'm thinking about `pod.spec.podIP / pod.spec.podIPs / service.spec.ClusterIP / service.spec.ClusterIPs / node.spec.nodeIP`, so I don't really think it's needed here. This component as we've already mentioned: should be even less user facing than any of those other fields, so I see an even lower need for it to be done here.

pkg/crd/cloudprivateipconfig/v1/doc.go
```diff
@@ -0,0 +1,4 @@
+// Package v1 contains API Schema definitions for the cloud v1 API group
```
tssurya, line 1: should this live here or in `openshift/api`? reference: there was a convo on slack around crds should live in api repo.

alexanderConstantinescu: Oh, could you link to the slack thread where the discussion was happening?

alexanderConstantinescu:
OK, it seems this is the wrong place for openshift CRDs. I am thus closing this PR and creating one in `openshift/api`

/close

PR #6, opened 2021-05-17
Cloud-network-config-controller implementation 

alexanderConstantinescu:
This PR implements the core control loops for managing the resource `CloudPrivateIPConfig`, supporting the feature: egress IP on cloud environments. 

Main aspects of this PR:

- Implements a node control loop: intended on retrieving the node's subnet information, as known by the cloud
- Implements a secret control loop: used for cloud credential secret rotation 
- Implements the control loop for `CloudPrivateIPConfig`

This controller has been manually tested on all clouds (Azure, AWS, GCP) and works as expected. Remaining integration with network plugins remains as well as a CNO PR for the deployment of the feature.


pkg/cloudprovider/aws.go
```diff
@@ -1,37 +1,247 @@
...
+		return err
+	}
+	mySession := session.Must(session.NewSession())
+	a.client = ec2.New(mySession, awsapi.NewConfig().WithCredentials(awscredentials.NewStaticCredentials(accessKey, secretKey, "")).WithRegion(a.region))
 	return nil
 }
 
 func (a *AWS) AssignPrivateIP(ip net.IP, node *corev1.Node) (interface{}, error) {
-	return nil, nil
+	instance, err := a.getInstance(node)
```
squeed, line 60: Can you add a basic docblock here?

pkg/cloudprovider/aws.go
```diff
@@ -1,37 +1,247 @@
...
+	if err != nil {
+		return nil, err
+	}
+	addIP := ip.String()
+	keepIPs := []*string{}
+	if utilnet.IsIPv6(ip) {
+		for _, assignedIPv6 := range instance.NetworkInterfaces[0].Ipv6Addresses {
```
squeed, line 67: Is it possible that NetworkInterfaces is an empty list? If so, this will panic

cmd/cloud-network-config-controller/main.go
```diff
@@ -49,70 +64,115 @@ func main() {
...
 	flag.StringVar(&kubeconfig, "kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
 	flag.StringVar(&masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 	flag.StringVar(&cloudProvider, "cloudprovider", "", "The cloud provider this component is running on.")
 	flag.StringVar(&cloudRegion, "cloudregion", "", "The cloud region the cluster is deployed in, this is explicitly required for talking to the AWS API.")
 	flag.StringVar(&secretName, "secret-name", "", "The cloud provider secret name - used for talking to the cloud API.")
 	flag.StringVar(&secretNamespace, "secret-namespace", "", "The cloud provider secret namespace - used for talking to the cloud API.")
+
+	// These are populate by the downward API
+	podNamespace = os.Getenv("POD_NAMESPACE")
+	podName = os.Getenv("POD_NAME")
```
squeed, line 175: You should exit 1 with a warning somewhere if these are empty string.

cmd/cloud-network-config-admission-controller/main.go
```diff
@@ -0,0 +1,156 @@
...
+	mux.Handle("/"+cloudprivateipconfigadmissioncontroller.CloudPrivateIPConfigResource.Resource, admitFuncHandler(admissionController))
+	server := &http.Server{
+		Addr:    ":8443",
+		Handler: mux,
+	}
+
+	klog.Fatal(server.ListenAndServeTLS(certPath, keyPath))
+}
+
+func doServeAdmitFunc(w http.ResponseWriter, r *http.Request, admit AdmissionControllerIntf) ([]byte, error) {
```
squeed, line 81: This function needs a docblock. What are the admission criteria?

cmd/cloud-network-config-admission-controller/main.go
```diff
@@ -0,0 +1,156 @@
+package main
```
squeed, line 1: It would be good if there was a README in this folder with a quick overview. Doesn't need to be too big, just a paragraph explaining what is being checked.

pkg/cloudprovider/gcp.go
```diff
@@ -13,25 +19,171 @@ const (
...
+	return instance, nil
+}
+
+// GCP Subnet URLs are defined as:
+// - https://www.googleapis.com/compute/v1/projects/project/regions/region/subnetworks/subnetwork
+// OR
+// - regions/region/subnetworks/subnetwork
+func (g *GCP) parseSubnet(subnetURL string) (string, string) {
+	subnetURLParts := strings.Split(subnetURL, "/")
+	return subnetURLParts[len(subnetURLParts)-3], subnetURLParts[len(subnetURLParts)-1]
```
squeed, line 179: Do a length check here, otherwise you'll panic.

pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
```diff
@@ -0,0 +1,498 @@
...
+			},
+		}
+		if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			cloudPrivateIPConfig, err = c.updateCloudPrivateIPConfigStatus(cloudPrivateIPConfig, status)
+			return err
+		}); err != nil {
+			return fmt.Errorf("Error updating CloudPrivateIPConfig: %s during delete operation, err: %v", name, err)
+		}
+		// This is a long running and blocking function call.
+		cloudErr := c.CloudProviderClient.WaitForResponse(cloudRequestObj)
```
squeed, line 230: Is it OK that we're blocked on other requests while this is running? Does it make sense to try and make this code thread-safe, and thus allow for multiple in-flight requests?

(I noticed that the AWS implementation does some checking that IPs are in a desired state; that might not work if running in parallel)

pkg/cloudprovider/aws.go
alexanderConstantinescu: I don't think so. The testing and research I've made on AWS seems to indicate that a VM will always have at least one network interface. I am adding a check anyways however.

cmd/cloud-network-config-admission-controller/main.go
alexanderConstantinescu: I added a doc bloc in this file for the `AdmissionFunc`, hopefully that is fine.

pkg/cloudprovider/gcp.go
alexanderConstantinescu: It might just be better if we panic if google decides to change their GCP API? At least like this it will be more "evident" to us

pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
alexanderConstantinescu: This code is thread-safe and does allow for multiple in-flight requests. `defaultWorkerThreadiness` is 10, which means we can process 10 simultaneous items off of the workqueue.

pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,131 @@
...
+	})
+	return controller
+}
+
+// syncHandler compares the actual state with the desired, and attempts to
+// converge the two. It then updates the Status block of the Node resource
+// with the current status of the resource.
+func (n *NodeController) SyncHandler(key string) error {
+	// Convert the key to a name (since Node is non-namespaced)
+	_, name, err := cache.SplitMetaNamespaceKey(key)
```
danwinship, line 76: for non-namespaced objects, `cache.MetaNamespaceKeyFunc()` just returns the name anyway... there's no splitting to be done here. (likewise in some other places)

pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		// The Node resource may no longer exist, in which case we stop processing.
+		if errors.IsNotFound(err) {
+			utilruntime.HandleError(fmt.Errorf("corev1.Node: '%s' in work queue no longer exists", key))
+			return nil
+		}
+		return fmt.Errorf("error retrieving corev1.Node from the API server, err: %v", err)
+	}
+	// If the node already has the annotation (ex: if we restart it is expected
+	// that the nodes would) we skip it. Subnets won't change and we are only
```
danwinship, line 91: If a node is deleted and recreated with a new subnet while the CNCC is not running... oh, then the new node wouldn't have the annotation. OK.

danwinship (commented):
(This is just for "Controller and cloud provider API"... more to come

pkg/cloudprovider/aws.go
```diff
@@ -1,37 +1,338 @@
...
+	CloudProvider
+	region string
+	client *ec2.EC2
+}
+
+// AWSWaitInput is the required input for the AWS EC2 Wait API call (WaitUntilInstanceRunning).
+// Unfortunately the API only handles equality assertion: so on delete we can't
+// specify and assert that the IP which is being removed is completely removed,
+// we are forced to do the inverse, i.e: assert that all IPs except the IP being
+// removed are there (which should anyways be the case). Hence we might risk
```
danwinship, line 38: That means the operation may fail if someone else modifies IPs at the same time we do, because the returned list will have either more or fewer IPs than we were expecting, right?

I don't know anything about the AWS APIs, but it seems weird that there is no better way to do this...

pkg/cloudprovider/aws.go
```diff
@@ -1,37 +1,338 @@
...
+// having a very small time window where the state is incorrect. This can be
+// problematic for update operation. The assumption here is that AWS EC2 will be
```
danwinship, line 40: I don't understand how this issue leads to "a very small time window where the state is incorrect" or why that's only a problem for updates.

pkg/cloudprovider/aws.go
```diff
@@ -1,37 +1,338 @@
...
+		if err != nil {
+			return nil, nil, fmt.Errorf("error: unable to parse IPv4 subnet, err: %v", err)
+		}
+		v4Subnet = subnet
+	}
+
+	// I don't know what it means to have several IPv6 CIDR blocks defined for
+	// one subnet, specially given that you can only have one IPv4 CIDR block
+	// defined...¯\_(ツ)_/¯
+	// Let's just pick the first.
```
danwinship, line 259: I'm not sure that's safe.

I guess really you want the IPv6 subnet that contains the IPv6 node IP, right?

pkg/cloudprovider/gcp.go
```diff
@@ -1,37 +1,248 @@
...
+}
+
+/*
+GCP hard-codes the amount of alias IPs that can be assigned to a NIC to 10 -
+independently of IP family, so we need to retrive the amount of alias IPs
+already in use by default and subtract from 10. See:
+https://cloud.google.com/vpc/docs/quota#per_instance .
+
+Note: there is also a global "alias IP per VPC quota", but OpenShift clusters on
+GCP seem to have that value defined to 15,000. So we can skip that.
```
danwinship, line 192: This may be because of the instance types or zones or whatever that we're using, which may not be the ones that everyone else is using?

(Also, `//` for comment)

pkg/admissioncontroller/cloudprivateipconfig/cloudprivateipconfig_admission_controller.go
```diff
@@ -0,0 +1,55 @@
...
+		AdmissionController: admissioncontroller.AdmissionController{
+			CloudNetworkClient: cloudNetworkClient,
+			KubeClient:         kubeClient,
+		},
+	}
+}
+
+// AdmissionFunc is the implementation of AdmissionControllerIntf for the
+// CloudPrivateIPConfig resource. This controller admission criteria is: the
+// node specified in .spec.node must exist
```
danwinship, line 42: Why do we need to enforce that? Why not just have the config controller set a failed condition on it? The config controller needs to be able to deal with the existence of configs for non-existent nodes anyway, due to race conditions, restarts, etc.

And CloudPrivateIPConfigs should only be created by openshift-sdn/ovn-kubernetes anyway, not directly by the user, so we can just ensure that they only generate correct objects.

I don't think there's enough benefit to this admission controller to make it worth running it and maintaining the code.

pkg/cloudprovider/cloudprovider.go
```diff
@@ -76,7 +75,7 @@ type NodeEgressIPConfiguration struct {
 	Capacity  capacity `json:"capacity"`
 }
 
-func NewCloudProviderClient(cloudProvider, cloudRegion string) (CloudProviderIntf, error) {
+func NewCloudProviderClient(cloudProvider, cloudSecretOverride, cloudRegion string) (CloudProviderIntf, error) {
```
danwinship, line 78: logically, "provider, region, secret override" would make more sense

danwinship:
Overall this basically looks good

Makefile
```diff
@@ -1,4 +1,4 @@
 build:
-	CGO_ENABLED=0 GO111MODULE=on go build -mod vendor -o _output/bin/cloud-network-config-controller cmd/cloud-network-config-controller/cloud-network-config-controller.go
+	CGO_ENABLED=0 GO111MODULE=on go build -mod vendor -o _output/bin/cloud-network-config-controller cmd/cloud-network-config-controller/main.go
```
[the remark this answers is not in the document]
alexanderConstantinescu: Filed: https://issues.redhat.com/browse/SDN-2467

alexanderConstantinescu (commented):
Thanks for the reviews @danwinship / @martinkennelly ! I've addresses all comments minus everything mentioned on the admission controller, mainly because I agree with Dan that it's not worth maintaining the code given it's low admission criteria. I will push an update next week, but I thought I might as well post my comments so that you can comment, if needed.

pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
```diff
@@ -0,0 +1,506 @@
...
+	}
+
+	// This object will recursively be updated during this sync
+	cloudPrivateIPConfig, err := c.cloudPrivateIPConfigLister.Get(name)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// The object was deleted while we were processing the request,
+			// nothing more to do, the finalizer portion of this sync should
+			// have handled the last cleanup
+			utilruntime.HandleError(fmt.Errorf("CloudPrivateIPConfig: '%s' in work queue no longer exists", name))
```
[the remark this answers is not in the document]
danwinship: I mean, if it was going to be a best practice then we should add a version where you can just pass the message directly rather than having to create an error just so you can pass it to the function.

But `HandleError` is only interesting if you also use the `utilruntime` functions for overriding its behavior anyway. If you never call those, then it's just a long-winded way of calling `klog.Errorf`.

pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
```diff
@@ -0,0 +1,506 @@
...
+			if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				cloudPrivateIPConfig, err = c.updateCloudPrivateIPConfigStatus(cloudPrivateIPConfig, status)
+				return err
+			}); err != nil {
+				return fmt.Errorf("Error updating CloudPrivateIPConfig: %s during delete operation, err: %v", name, err)
+			}
+		}
+	}
+
+	if nodeToAdd != "" {
```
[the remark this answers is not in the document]
danwinship: Multiple updates to the same IP. eg:

1. (IP1 is assigned to Node1)
2. Node1 goes down, SDN assigns IP1 to Node2
3. CNCC starts deleting IP1 from Node1...
4. Node2 goes down, SDN assigns IP1 to Node 3
5. CNCC finishes deleting IP1 from Node1
6. CNCC adds IP1 to Node2
7. CNCC deletes IP1 from Node2
8. CNCC adds IP1 to Node3

In theory, it should not do steps 6 and 7, but with the current code it will.

This may not be that big a deal.

pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
```diff
@@ -0,0 +1,506 @@
...
+	cloudPrivateIPConfigCopy := cloudPrivateIPConfig.DeepCopy()
+	cloudPrivateIPConfigCopy.Status = *status
+	return c.cloudNetworkClientset.CloudV1().CloudPrivateIPConfigs().UpdateStatus(context.TODO(), cloudPrivateIPConfigCopy, metav1.UpdateOptions{})
+}
+
+// updateCloudPrivateIPConfig copies and updates the provided object and returns
+// the new object. The return value can be useful for recursive updates
+func (c *CloudPrivateIPConfigController) updateCloudPrivateIPConfig(cloudPrivateIPConfig *cloudnetworkv1.CloudPrivateIPConfig) (*cloudnetworkv1.CloudPrivateIPConfig, error) {
+	cloudPrivateIPConfigCopy := cloudPrivateIPConfig.DeepCopy()
+	return c.cloudNetworkClientset.CloudV1().CloudPrivateIPConfigs().Update(context.TODO(), cloudPrivateIPConfigCopy, metav1.UpdateOptions{})
```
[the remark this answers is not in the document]
danwinship: That "it" is a little vague I guess... Update takes a representation of a cloudPrivateIPConfig, and updates the cloudPrivateIPConfig on the apiserver to match the provided representation.

pkg/controller/node/node_controller.go
```diff
@@ -0,0 +1,131 @@
...
+	// interested in conveying the default assignment capacity that the node had
+	// when it started existing. It's up to the network plugin to track how much
+	// capacity it has left depending on the assignments it performs.
+	annotations := node.GetAnnotations()
+	if _, ok := annotations[nodeEgressIPConfigAnnotationKey]; ok {
```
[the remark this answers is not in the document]
danwinship: The getters are part of the `metav1.Object` interface, so that in theory you can have APIs that work on any API object type, but this almost never happens in practice (at least, outside of the apiserver).

pkg/cloudprovider/cloudprovider.go
```diff
@@ -0,0 +1,99 @@
...
+	// VM instance corresponding to the corev1.Node provided on the cloud the
+	// cluster is deployed on. NOTE: this operation is only performed against
+	// the first network interface defined for the VM.
+	ReleasePrivateIP(ip net.IP, node *corev1.Node) (interface{}, error)
+	// WaitForResponse runs a long function running call waiting for the cloud's
+	// response to the previously called Assign/ReleasePrivateIP. If it timeouts
+	// or encounters an error, that error is then returned. The function
+	// argument accepts whatever resource needed by the cloud to determine the
+	// status of the in-flight request.
+	WaitForResponse(interface{}) error
```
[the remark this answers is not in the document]
danwinship: I mean that `AssignPrivateIP` and `RemovePrivateIP` are asynchronous calls, with `WaitForResponse` as their "finish" function. And there doesn't seem to be any need for this; sure, the underlying cloud APIs are asynchronous, but you could just have each cloud implementation do the asynchronous waiting itself, and not return from `AssignPrivateIP` / `RemovePrivateIP` until you've gotten back the result.

pkg/cloudprovider/azure.go
```diff
@@ -1,37 +1,294 @@
...
 
 func (a *Azure) AssignPrivateIP(ip net.IP, node *corev1.Node) (interface{}, error) {
-	return nil, nil
+	instance, err := a.getInstance(node)
+	if err != nil {
+		return nil, err
+	}
+	networkInterface := network.Interface{}
+	for _, netif := range *instance.NetworkProfile.NetworkInterfaces {
+		if *netif.Primary {
```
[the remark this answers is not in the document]
danwinship: well, if we're assigning egress IPs to it then it must be a `Node` with status `Ready`, which definitely _strongly implies_ the existence of a NIC :slightly_smiling_face: .

Maybe just run through the code and for each deref, think about whether it could plausibly be `nil`, and make sure there are checks in those cases, but not in the "can't happen" cases.

pkg/admissioncontroller/cloudprivateipconfig/cloudprivateipconfig_admission_controller.go
danwinship: yeah, cluster-admin is always supposed to be able to create/edit any object

pkg/cloudprovider/cloudprovider.go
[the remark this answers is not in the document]
alexanderConstantinescu: Hmm, I see now: indeed that is a good simplification. I will update it

pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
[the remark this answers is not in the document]
alexanderConstantinescu: Not sure about that, we're using a queue informer and I thought that the delta FIFO will make sure the events are coalesced so that we only get the latest update, i.e: the following:

1. (IP1 is assigned to Node1)
2. Node1 goes down, SDN assigns IP1 to Node2
3. CNCC starts deleting IP1 from Node1...
4. Node2 goes down, SDN assigns IP1 to Node 3
5. CNCC finishes deleting IP1 from Node1
6. CNCC adds IP1 to Node3, because it gets the latest update on the object

Might be worth an experiment. I will see if I can formulate this as a unit test, otherwise I will manually test it out by adding a sleep in the sync.

pkg/admissioncontroller/cloudprivateipconfig/cloudprivateipconfig_admission_controller.go
alexanderConstantinescu: OK, I am dropping the admission controller.

pkg/controller/secret/secret_controller.go
```diff
@@ -0,0 +1,131 @@
...
+	corelisters "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog/v2"
+
+	controller "github.com/openshift/cloud-network-config-controller/pkg/controller"
+)
+
+var (
+	// secretControllerAgentType is the Secret controller's dedicated resource type
+	secretControllerAgentType reflect.Type = reflect.TypeOf(&corev1.Secret{})
```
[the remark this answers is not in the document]
martinkennelly: Major nit - feel free to ignore: I meant no need defining `reflect.Type` so `secretControllerAgentType reflect.Type = reflect.TypeOf(&corev1.Secret{})` -> `secretControllerAgentType = reflect.TypeOf(&corev1.Secret{})`

pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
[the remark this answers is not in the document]
alexanderConstantinescu: I've just tested this and what I stated seems to be correct: the updates are coalesced, so we will only get the latest update in case we are slow and happen to have multiple updates getting "stacked up".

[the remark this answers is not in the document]
alexanderConstantinescu: Never mind what I said, I was confused and wrote a dubious test which didn't verify anything of importance. I am not sure this might be very relevant in reality, but it's a real corner case to take into account. 

I need to think about it, but I suspect the solution is to XOR delete and add for update operations so that they won't be attempted during the same sync, that should have the example above solved by the update events being coalesced. I just need to think about other corner cases which might be impacted negatively by this, and make sure that is handled.

pkg/cloudprovider/aws.go
alexanderConstantinescu: I've been thinking about this and I am not sure that makes more sense. What if the node has multiple IPv6 addresses? If it has one it probably has one or multiple CIDR blocks. 

Now, if there are multiple CIDR blocks we probably need to redefine the annotation, since it currently only assumes one-to-one relation between IP family subnets and NIC. Since all of this is very speculative, conceptual and strange at this point (since I am not able to bring up a IPv6 / dual-stack cluster on AWS and verify this): I think it's fine to just take the first IPv6 CIDR block if there are multiple defined, and not redefine the annotation for now.

alexanderConstantinescu:
@danwinship / @martinkennelly : thanks for the reviews. I've been updating the PR and think I have something close to a finish product now. Recent changes include:

- Incremental sync, i.e we only process and ADD / DEL per sync not both. 
- Taken into account all of your comments
- Removed the admission controller

cmd/cloud-network-config-controller/main.go
```diff
@@ -0,0 +1,184 @@
...
+				if err != nil {
+					klog.Fatalf("Error building cloudnetwork clientset: %s", err.Error())
+				}
+
+				cloudProviderClient, err := cloudprovider.NewCloudProviderClient(cloudProvider, cloudSecretOverride, cloudRegion)
+				if err != nil {
+					klog.Fatal("Error building cloud provider client, err: %v", err)
+				}
+
+				kubeInformerFactory := kubeinformers.NewSharedInformerFactoryWithOptions(kubeClient, time.Second*30, kubeinformers.WithNamespace(controllerNamespace))
```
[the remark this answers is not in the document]
alexanderConstantinescu: I finally got to the bottom of this: when the re-sync period kicks in the informer's `UpdateFunc` gets called. I've set this value to 2 minutes today and the only controller actually reacting to the `UpdateFunc` is the cloudprivateipconfig controller and the secret controller. The cost of the resync is however very minimal. Both controllers wont enqueue the object for sync unless there are any changes we care about, so it's not very costly

pkg/cloudprovider/aws.go
```diff
@@ -1,32 +1,342 @@
...
+				NetworkInterfaceId: networkInterface.NetworkInterfaceId,
+				Ipv6Addresses:      keepIPs,
+			}
+			_, err = a.client.AssignIpv6Addresses(&input)
+			if err != nil {
+				return err
+			}
+			return a.waitForCompletion(node, awsapi.StringValueSlice(keepIPs), false)
+		}
+		for _, assignedIPv4 := range networkInterface.PrivateIpAddresses {
```
danwinship, line 82: I'd do `} else {` here so the IPv4 code is indented at the same level as the IPv6 code... (likewise in ReleasePrivateIP)

pkg/cloudprovider/aws.go
```diff
@@ -1,32 +1,342 @@
...
+		if err != nil {
+			return err
+		}
+		return a.waitForCompletion(node, awsapi.StringValueSlice(keepIPs), false)
+	}
 	return nil
 }
 
+// ReleasePrivateIP un-assigns the IP address from the node by re-assigning the
+// existing IP addresses - the one to be removed.
```
danwinship, line 103: This is no longer accurate, right? You appear to be using a delete function.

pkg/cloudprovider/aws.go
```diff
@@ -1,32 +1,342 @@
...
+					deleteIPs = append(deleteIPs, assignedIPv6.Ipv6Address)
+				}
+			}
+			if len(deleteIPs) == 0 {
+				return NonExistingIPError
+			}
+			input := ec2.UnassignIpv6AddressesInput{
+				NetworkInterfaceId: networkInterface.NetworkInterfaceId,
+				Ipv6Addresses:      deleteIPs,
+			}
```
danwinship, line 129: Do you really need to compute `deleteIPs`? Can't you just do
```
			deleteIP := ip.String()
			input := ec2.UnassignIpv6AddressesInput{
				NetworkInterfaceId: networkInterface.NetworkInterfaceId,
				Ipv6Addresses:      [&deleteIP],
			}
```

and in the unlikely event that it's `NonExistingIPError` you can figure that out from the AWS error?

pkg/cloudprovider/aws.go
```diff
@@ -1,32 +1,342 @@
...
+
+// Unfortunately the AWS API (WaitUntilInstanceRunning) only handles equality
+// assertion: so on delete we can't specify and assert that the IP which is
+// being removed is completely removed, we are forced to do the inverse, i.e:
+// assert that all IPs except the IP being removed are there (which should
+// anyways be the case). Hence, use our own poller which verifies on ADD that
+// all IPs that we add are assigned to the node, and on DEL that all IPs being
+// removed have been completely removed from the node.
+func (a *AWS) waitForCompletion(node *corev1.Node, ips []string, deleteOp bool) error {
+	return wait.PollImmediate(time.Second*2, time.Minute, func() (done bool, err error) {
```
danwinship, line 203: it's generally preferred to use `wait.ExponentialBackoff`

pkg/cloudprovider/aws.go
```diff
@@ -1,32 +1,342 @@
...
+			for _, assignedIPv4 := range instance.NetworkInterfaces[0].PrivateIpAddresses {
+				if assignedIP := net.ParseIP(*assignedIPv4.PrivateIpAddress); assignedIP != nil {
+					assignedIPs = append(assignedIPs, assignedIP.String())
+				}
+			}
+		}
+		if !deleteOp {
+			return sets.NewString(assignedIPs...).HasAll(ips...), nil
+		}
+		return !sets.NewString(assignedIPs...).HasAny(ips...), nil
```
danwinship, line 226:
```
if deleteOp {
        return !sets.NewString(assignedIPs...).HasAny(ips...), nil
} else {
        return sets.NewString(assignedIPs...).HasAll(ips...), nil
}
```

the way it is now suggests that `!deleteOp` is a special case to be handled first, and then `return !sets.NewString(assignedIPs...).HasAny(ips...), nil` is the default behavior. But neither is default or special-case. They're equal.

pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
```diff
@@ -0,0 +1,437 @@
...
+// computeOp decides on what needs to be done given the state of the object.
+func (c *CloudPrivateIPConfigController) computeOp(cloudPrivateIPConfig *cloudnetworkv1.CloudPrivateIPConfig) (string, string) {
+	// Delete if the deletion timestamp is set and we still have our finalizer listed
+	if !cloudPrivateIPConfig.DeletionTimestamp.IsZero() && controllerutil.ContainsFinalizer(cloudPrivateIPConfig, cloudPrivateIPConfigFinalizer) {
+		return "", cloudPrivateIPConfig.Status.Node
+	}
+	// Update if status and spec are different
+	if cloudPrivateIPConfig.Spec.Node != cloudPrivateIPConfig.Status.Node && cloudPrivateIPConfig.Status.Node != "" {
+		return "", cloudPrivateIPConfig.Status.Node
+	}
```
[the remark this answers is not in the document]
alexanderConstantinescu: > What tells us to re-add it in the next sync?

When the current status assignment finally and successfully is deleted: we'll finish by updating the status in the delete code to `""`. Given that the object is updated, our sync loop will be triggered again and given the new status on the object will reach the add code because `cloudPrivateIPConfig.Status.Node == ""`

pkg/cloudprovider/azure.go
```diff
@@ -1,37 +1,354 @@
...
+		return nil, err
+	}
+	networkInterfaces, err := a.getNetworkInterfaces(instance)
+	if err != nil {
+		return nil, err
+	}
+	// Perform the operation against the first interface listed, which will be
+	// the primary interface (if it's defined as such) or the first one returned
+	// following the order Azure specifies.
+	for _, networkInterface := range networkInterfaces {
```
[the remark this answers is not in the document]
alexanderConstantinescu: The Azure code was a bit off: I just realized it had the network interface validation baked inside `getInstance`. I moved that to `getNetworkInterfaces` as to have it be more explicit that when we reach this part of the code: we are guaranteed to have at least one NIC. 

The reason I chose to write the code like this was to minimize the change in case we were ever thinking to deal with more than just the primary NIC.

pkg/cloudprovider/azure.go
```diff
@@ -1,37 +1,354 @@
...
+func (a *Azure) getNetworkInterface(id string) (network.Interface, error) {
+	ctx, cancel := context.WithTimeout(a.ctx, defaultAzureOperationTimeout)
+	defer cancel()
+	return a.networkClient.Get(ctx, a.resourceGroup, getNameFromResourceID(id), "")
+}
+
+// This is what the subnet ID looks like on Azure:
+// 	ID: "/subscriptions/d38f1e38-4bed-438e-b227-833f997adf6a/resourceGroups/ci-ln-wzc83kk-002ac-qcghn-rg/providers/Microsoft.Network/virtualNetworks/ci-ln-wzc83kk-002ac-qcghn-vnet/subnets/ci-ln-wzc83kk-002ac-qcghn-worker-subnet"
+func (a *Azure) getVirtualNetworkNameFromSubnetID(subnetID string) (string, error) {
+	providerData := splitURI(subnetID)
```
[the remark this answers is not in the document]
alexanderConstantinescu: `splitPath`?

pkg/cloudprovider/aws.go
alexanderConstantinescu: > and in the unlikely event that it's NonExistingIPError you can figure that out from the AWS error?

I don't know, presumably the API should handle it and not return an error if you provide it an empty list of IPs, but I am not sure. I'd prefer we merge the code as is (since I've done some basic testing on all platforms verifying that this works), and I don't think we have time to investigate the cloud API at this point.

pkg/cloudprovider/aws.go
alexanderConstantinescu: I thought it would be more straightforward to have a fixed frequency for polling with a specified deadline. `ExponentialBackoff` would make things a lot less non-obvious w.r.t how we are polling.

pkg/cloudprovider/aws.go
alexanderConstantinescu: Indeed it's not. I have changed the doc bloc.

pkg/cloudprovider/aws.go
```diff
@@ -1,32 +1,342 @@
...
+	if err != nil {
+		return err
+	}
+	networkInterfaces, err := a.getNetworkInterfaces(instance)
+	if err != nil {
+		return err
+	}
+	// Perform the operation against the first interface listed following the
+	// order AWS specifies.
+	for _, networkInterface := range networkInterfaces {
```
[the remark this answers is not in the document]
alexanderConstantinescu: OK, I will change this part across all cloud providers.

pkg/cloudprovider/azure.go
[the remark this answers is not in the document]
alexanderConstantinescu: I changed it explicitly dealing with only the first item in the slice:

pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
```diff
@@ -0,0 +1,452 @@
...
+	})
+	return updatedCloudPrivateIPConfig, err
+}
+
+// getCloudPrivateIPConfig retrieves the object from the API server
+func (c *CloudPrivateIPConfigController) getCloudPrivateIPConfig(name string) (*cloudnetworkv1.CloudPrivateIPConfig, error) {
+	ctx, cancel := context.WithTimeout(c.ctx, controller.ClientTimeout)
+	defer cancel()
+	// This object will repeatedly be updated during this sync, hence we need to
+	// retrieve the object from the API server as supposed to the informer cache
```
danwinship, line 397: "as opposed"

PR #11, opened 2021-12-16
Add .ci-operator.yaml

alexanderConstantinescu:
This adds a `.ci-operator.yaml` following the specification of such from https://docs.ci.openshift.org/docs/architecture/ci-operator/#build-root-image 

This is a precursor to having the bot automatically bump the test images defined at: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cloud-network-config-controller/openshift-cloud-network-config-controller-master.yaml#L11-L15 which will help when bumping vendored modules which require a go bump

/assign @danwinship

### DIFF
--- a/.ci-operator.yaml
+++ b/.ci-operator.yaml
@@ -1,0 +1,4 @@
+build_root_image:
+  name: release
+  namespace: openshift
+  tag: rhel-8-release-golang-1.17-openshift-4.10


danwinship:
/lgtm
/approve

I don't know what this means, but it matches our other modules

alexanderConstantinescu:
> I don't know what this means, but it matches our other modules

CI tests (ex: `ci/prow/unit`) are executed in a container, the container image is specified in: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cloud-network-config-controller/openshift-cloud-network-config-controller-master.yaml#L11-L15 (which means that this is currently done in a go 1.15 based container). We don't want to manually bump that container image version for every go release, and sometimes we don't have to, but certain versions of code are coupled to a go version (Kubernetes 1.23 requires go 1.17). The goal of doing this is thus to replace that part of the release config and instead specify something along the lines of what we're doing for openshift-sdn: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/sdn/openshift-sdn-master.yaml#L11-L12. Doing that will have the config file reference this `.ci-operator.yaml` located in the corresponding repo, and will also have the openshift-bot bump the `.ci-operator.yaml` for every OpenShift release (thus creating a more aligned test container w.r.t the code that we are vendoring).

PR #13, opened 2021-12-23
Bug 2035093: AWS: Allow using auth methods other than static credentials

alvaroaleman:
This changes the AWS credentials loading to load the config file rather
than expecting static credentials files to be there. In the default case
of selfhosted OCP, this ends up doing the same as the config file
contains the static credentials but it will also work if a different
auth method like STS is defined. This allows the cncc to work in Hypershift.

### DIFF
--- a/pkg/cloudprovider/aws.go
+++ b/pkg/cloudprovider/aws.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	awsapi "github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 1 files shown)


alexanderConstantinescu (approved):
/lgtm

squeed:
FWIW, now that I've done a bit more research, I believe there is also still a bug here.

The CCO may provide either `credentials` or `access_key` fields, especially when we use user-provisioned secrets.

I wonder if this is also the case for other cloud providers...

PR #14, opened 2022-01-06
Bug 2034460: cloud-provider: handle advanced AWS and Azure configurations 

squeed:
It turns out connecting to AWS and Azure isn't so simple. This change
does the following:

- Reads the AWS endpoint override from the Infrastructure object
- Pulls in the AWS CA bundle from the kube-cloud-config ConfigMap
- Reads the Azure Environment and URL overrides from the Infrastructure
  object

Fixes: 2034460


pkg/cloudprovider/cloudprovider.go
```diff
@@ -87,9 +106,20 @@ type NodeEgressIPConfiguration struct {
 	Capacity  capacity `json:"capacity"`
 }
 
-func NewCloudProviderClient(platformType, platformRegion, secretOverride string) (CloudProviderIntf, error) {
+func NewCloudProviderClient(kubeClient *kubernetes.Clientset, configClient configclientset.Interface, secretOverride string) (CloudProviderIntf, error) {
 	var cloudProviderIntf CloudProviderIntf
 
+	// Get the cloud provider configuration from the API
+	infra, err := configClient.ConfigV1().Infrastructures().Get(context.Background(), "cluster", v1.GetOptions{})
```
alexanderConstantinescu, line 113: We can't do it like this. There can be no hard dependency from the CNCC on other OpenShift components, this is because upstream OVN-Kubernetes needs to be able to deploy this component and have it work by only providing the secrets required to talk to the cloud API, as mentioned in the enhancement proposal: https://github.com/alexanderConstantinescu/enhancements/blob/3231f1103145f5ef5b7b88bc6aab811b834088b0/enhancements/network/cloud-egress-ip.md#openshift-vs-upstream-project . Upstream should not to deploy other OpenShift components to have this one work. Hence we need a default behavior intended for reading "simple credenentials/secrets" and then override that in case we deploy something "obscure only OpenShift does" type of thing.

alexanderConstantinescu: This is also the reason the `platform-type` / `platform-region` etc is passed to this component though its deployment stanza

alexanderConstantinescu (changes requested):

squeed:
OK, updated to just get configuration from cli args. Will need a CNO PR too, will file shortly.

squeed:
@alexanderConstantinescu PTAL, thanks

pkg/cloudprovider/azure.go
```diff
@@ -63,17 +64,27 @@ func (a *Azure) initCredentials() error {
 		return err
 	}
 
-	a.vmClient = compute.NewVirtualMachinesClient(subscriptionID)
+	// Pick the Azure "Environment", which is just a named set of API endpoints.
+	if a.cfg.APIOverride != "" {
+		a.env, err = azure.EnvironmentFromURL(a.cfg.APIOverride)
+	} else {
+		a.env, err = azure.EnvironmentFromName(a.cfg.AzureEnvironment)
+	}
```
alexanderConstantinescu, line 72: So there always needs to be either an `APIOverride` or `AzureEnvironment` ?

cmd/cloud-network-config-controller/main.go
```diff
@@ -171,14 +169,17 @@ func init() {
...
-	flag.StringVar(&secretOverride, "secret-override", "", "The cloud provider secret location override, useful when running this component locally against a cluster")
+	flag.StringVar(&platformCfg.PlatformType, "platform-type", "", "The cloud provider platform type this component is running on.")
+	flag.StringVar(&platformCfg.Region, "platform-region", "", "The cloud provider platform region the cluster is deployed in, required for AWS")
+	flag.StringVar(&platformCfg.APIOverride, "platform-api-url", "", "The cloud provider API URL to use (instead of whatever default).")
+	flag.StringVar(&platformCfg.CredentialDir, "secret-override", "/etc/secret/cloudprovider", "The cloud provider secret location override, useful when running this component locally against a cluster")
+	flag.StringVar(&platformCfg.AzureEnvironment, "platform-azure-environment", "AzurePublicCloud", "The Azure environment name, used to select API endpoints")
+	flag.StringVar(&platformCfg.AWSCAOverride, "platform-aws-ca-override", "", "Path to a separate CA bundle to use when connecting to the AWS api")
```
alexanderConstantinescu, line 177: Nit:  uppercase "api" in the description.

alexanderConstantinescu (commented):
/hold

Until the CNO PR is merged. 

Besides that I only have some minor comments, this looks good in general. Thanks!

pkg/cloudprovider/azure.go
squeed: Kind of, there is a default Azure environment specified. But Azure makes liberal use of the "environment" -- you even need to change it if you're using the Azure Germany cloud.

squeed:
/hold cancel
This needs to merge first, then the CNO change can go in (so it can use the new cli flags)

alexanderConstantinescu:
> This needs to merge first, then the CNO change can go in (so it can use the new cli flags)

Is it though? `AzureEnvironment` or `APIOverride` are required on Azure as things stand with this PR...yet the CNO doesn't provide these arguments. We will break all OpenShift CI jobs on Azure if we merge this before the CNO one, right?

alexanderConstantinescu:
It's required given this https://github.com/openshift/cloud-network-config-controller/pull/14#discussion_r781239078 right?

This would make it "not required"

```go
        if a.cfg.APIOverride != "" {
		a.env, err = azure.EnvironmentFromURL(a.cfg.APIOverride)
	} else if a.cfg.AzureEnvironment != "" {
		a.env, err = azure.EnvironmentFromName(a.cfg.AzureEnvironment)
	}
```

Well, I am assuming `azure.EnvironmentFromName` will fail for empty strings, maybe that is not the case?

I don't see any problems with AWS, so it should probably pass this repo's CI.

squeed:
huh, terraform wierdness:

```
level=error msg=When applying changes to module.vpc.aws_subnet.public_subnet[1], provider
level=error msg="openshift/local/aws" produced an unexpected new value: Root resource was
level=error msg=present, but now absent.
level=error
level=error msg=This is a bug in the provider, which should be reported in the provider's own
level=error msg=issue tracker. 
```

/retest

squeed:
/hold
can't get Azure to come up w/ cluster-bot

squeed:
argh
> level=error
level=error msg=Error: creating/updating Public Ip Address: (Name "ci-ln-m41tvj2-1d09d-4cr94-bootstrap-pip-v4" / Resource Group "ci-ln-m41tvj2-1d09d-4cr94-rg"): network.PublicIPAddressesClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: Code="PublicIPCountLimitReached" Message="Cannot create more than 100 public IP addresses for this subscription in this region." Details=[]
level=error

squeed:
/hold cancel
azure works fine

alexanderConstantinescu (approved):
/lgtm

PR #16, opened 2022-01-14
Bug 2034460: Update Azure Authentication for Different Environments

patrickdillon:
During my testing of https://bugzilla.redhat.com/show_bug.cgi?id=2034460 Azure Stack authentication failed.

The authorizer function is hardcoded to the public cloud: https://github.com/openshift/cloud-network-config-controller/blob/05c585b5f343d4f66f8dd29bb185776afbb793f2/pkg/cloudprovider/azure.go#L350-L360

I'm still in process of testing this PR so consider it a WIP, and also I think it may change the authentication method, so feedback is welcome.


patrickdillon:
I opened openshift/release#25299 to add azurestack tests. They are not the most stable tests, but I believe they could help in testing for installation here.

squeed:
thanks!
/approve
/lgtm
/override ci/prow/e2e-gcp
/bugzilla refresh

PR #17, opened 2022-01-17
Bug 2035167: fix cloudprivateipconfig sync on delete 

alexanderConstantinescu:
This PR fixes an issue experienced while processing deletes whereby in certain corner cases we could end up with a failed cloud release but a successful status update just before. The status update unset the `.status.node` field so on any succeeding sync we never retried the delete again, leading to an un-reconcilable state. 

It also fixes a spurious issue during adds, namely: if  `AssignPrivateIP` fails we unset the node in the status, this caused additional en-queuing of the same object by our `UpdateFunc`. If `AssignPrivateIP` fails we will still re-sync the object and re-attempt the add (since the `.status.conditions[0].status != True`), doing this however avoids both erroneously enqueuing the object and confusing logs.  

It also attempts improving the log message as to make them easier to discern when reading the logs, as well as using `getCloudPrivateIPConfig` when updating the CloudPrivateIPConfig during our sync (and hence using one context per call). 

/assign @danwinship

### DIFF
--- a/pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
+++ b/pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net"
@@ -16,6 +17,7 @@ import (
 	cloudprovider "github.com/openshift/cloud-network-config-controller/pkg/cloudprovider"
 	controller "github.com/openshift/cloud-network-config-controller/pkg/controller"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	coreinformers "k8s.io/client-go/informers/core/v1"
 	"k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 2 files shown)


pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
```diff
@@ -353,21 +353,24 @@ func (c *CloudPrivateIPConfigController) SyncHandler(key string) error {
...
-
 		// Get the latest item from the store before updating. We risk updating
 		// and overriding any client changes if we don't get the latest state of
 		// the object and update the status on that item. Ex: if we are in the
 		// middle of updating the status on item v1, after an update to the
 		// object v2 has been performed we'll override the changes made to v2
 		// back to v1 when updating below (because cloudPrivateIPConfig above
 		// will have the old state before any potential client changes).
-		cloudPrivateIPConfigLatest, err := c.cloudNetworkClient.CloudV1().CloudPrivateIPConfigs().Get(ctx, cloudPrivateIPConfig.Name, metav1.GetOptions{})
+		cloudPrivateIPConfigLatest, err := c.getCloudPrivateIPConfig(cloudPrivateIPConfig.Name)
```
danwinship, line 363: This doesn't seem right to me; you should pass this function's context down to `getCloudPrivateIPConfig`

alexanderConstantinescu: Why is that needed? `getCloudPrivateIPConfig` uses the global timeout defined for all kube client calls and can initiate it's own context based on that.

PR #18, opened 2022-01-18
Bug 2041882: gcp: retrieve project from resources, not from the credentials

squeed:
It turns out that GCP, when in "workload identity" mode, doesn't provide the project in the credentials file. The  easy fix: extract it from the node objects themselves.

### DIFF
--- a/pkg/cloudprovider/gcp.go
+++ b/pkg/cloudprovider/gcp.go
@@ -1,9 +1,9 @@
 package cloudprovider
 
 import (
-	"encoding/json"
 	"fmt"
 	"net"
+	"net/url"
 	"strings"
 
 	google "google.golang.org/api/compute/v1"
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/gcp.go
```diff
@@ -255,6 +237,28 @@ func (g *GCP) getNetworkInterfaces(instance *google.Instance) ([]*google.Network
...
+func splitGCPNode(node *corev1.Node) (project, zone, instance string, err error) {
+	u, err := url.Parse(node.Spec.ProviderID)
+	if err != nil {
+		err = fmt.Errorf("failed to parse node %s provider id %s: %w", node.Name, node.Spec.ProviderID, err)
+		return
+	}
+	project = u.Host
+	parts := strings.SplitN(u.Path, "/", 3)
+	if len(parts) != 3 {
+		err = fmt.Errorf("failed to parse node %s provider id %s: expected two path components", node.Name, node.Spec.ProviderID)
```
alexanderConstantinescu, line 254: Could it be that we panic below depending on how long `parts` is once we error here? Might be safer to just put a `return` right after the error

alexanderConstantinescu (commented):
Turns out credentials are not _that_ easy....

I just had a small comment, might be nothing in the end, but waiting for your answer.

pkg/cloudprovider/gcp.go
squeed: oh crap. Forgot the return. Very good catch!!

squeed:
> Turns out credentials are not _that_ easy....

I know! What a mess!

squeed:
@alexanderConstantinescu updated, thanks!

alexanderConstantinescu (approved):
/lgtm

Thanks for all the help on this!

PR #19, opened 2022-01-21
Bug 2035167: Revert to using `UpdateStatus`

alexanderConstantinescu:
`Patch` operations do not error, but do not work either against subresources like status when running against a real API server...using a `fakeClient` things work just fine and the status is updated :1st_place_medal: hence why our tests were indicating good results for #17 

/assign @danwinship

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,20 @@
 build:
 	CGO_ENABLED=0 GO111MODULE=on go build -mod vendor -o _output/bin/cloud-network-config-controller cmd/cloud-network-config-controller/main.go
 test:
-	go test ./... -count=1 -race
+	# This is commenting out the racy tests. The test file:
+	# cloudprivateipconfig_controller_racy_test.go has the following go build
+	# tag: "// +build race", which means it won't run if -race isn't provided to
+	# "go test", hence why commenting out the test execution below with the
+	# -race flag, effectively comments out the tests. The racy tests are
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,5 +1,16 @@
 build:
 	CGO_ENABLED=0 GO111MODULE=on go build -mod vendor -o _output/bin/cloud-network-config-controller cmd/cloud-network-config-controller/main.go
 test:
-	go test ./... -count=1 -race
+	# This is commenting out the racy tests. This is because the racy tests are
```
danwinship, line 4: I don't understand this. You aren't commenting out any tests, you're just telling go to not warn you that your tests have race conditions in them. But regardless of whether you let go warn you about it or not, the racy tests will still cause unit test flakes...

(Also, typo in the commit message: "Disbale")

alexanderConstantinescu: The test file containing the racy tests have this defined: 
https://github.com/openshift/cloud-network-config-controller/blob/df7a65440de18c14730605e3a1662399174b28dc/pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller_racy_test.go#L1
So it will only run for `go test -race`. Hence by commenting that part out from the Makefile, I am effectively "commenting out the test"

Noted for the spelling mistake, will update now.

alexanderConstantinescu: I could maybe re-phrase this to make it clearer.

danwinship:
ahhh
/lgtm

PR #26, opened 2022-03-03
Bug 2060334: Fix Azure VNET lookup when the NIC's subnet is in a different resource group

darthhexx:
In Azure cluster setups where the network resources, VNET and subnets, are in a different resource group to the cluster resources the VNET lookup fails.

```bash
$ oc get infrastructures.config.openshift.io cluster  -ojsonpath='{.status.platformStatus}' | jq
{
  "azure": {
    "cloudName": "AzurePublicCloud",
    "networkResourceGroupName": "v4-eastus",
    "resourceGroupName": "aro-dnewman-test-4-10"
  },
  "type": "Azure"
}
```
The error message that is being logged is: `The Resource 'Microsoft.Network/virtualNetworks/dev-vnet' under resource group 'aro-dnewman-test-4-10' was not found`, because it is actually in the 'v4-eastus' resource group.

The PR addresses the issue by parsing and using the NIC subnet's resource group instead of the global `azure_resourcegroup`.


squeed:
@darthhexx this should go in shortly - I assume this should be cherry-picked to 4.10?

darthhexx:
@squeed yes, ideally it should be cherry-picked into 4.10 to ensure we have the expected behaviour in ARO.

PR #27, opened 2022-03-03
Bug 2060361: Fix Unable to enumerate NICs due to a missing 'primary' field due to security restrictions

darthhexx:
If the security settings on a NIC prevent the enumeration of the NetworkInterfaceReferenceProperties to populate Primary field, then the NIC is not processed, leading to the 'cloud.network.openshift.io/egress-ipconfig' annotation not being applied to a node.

In order to overcome this limitation, this PR contains 2 commits to address the issues we are seeing on master VM nodes in ARO.

**Option 1** (commit e9a46a0): Does not rely on the Primary field being present to select the NIC, because there is only one to choose from.
**Option 2** (commit d9ddd33): Selects the first NIC in the list when the lookup based on the primary field has failed. This will have no side effects for VMs with only a single NIC, which is what we have in ARO. However, in multi-NIC scenarios it is a compromise since we do not _really_ know which one we should be picking.

At the very least we should go with the first commit in this PR to address the permissions issue on master nodes in ARO clusters, but either would work.

### DIFF
--- a/pkg/cloudprovider/azure.go
+++ b/pkg/cloudprovider/azure.go
@@ -279,6 +279,16 @@ func (a *Azure) getNetworkInterfaces(instance *compute.VirtualMachine) ([]networ
 		return nil, NoNetworkInterfaceError
 	}
 	networkInterfaces := []network.Interface{}
+	// When the VM has only a single NIC, don't rely on the presence of the Primary field, which is populated by
+	// NetworkInterfaceReferenceProperties, because it is not populated in some cases due to security restrictions
+	if len(*instance.NetworkProfile.NetworkInterfaces) == 1 {
+		intf, err := a.getNetworkInterface(*(*instance.NetworkProfile.NetworkInterfaces)[0].ID)
+		if err != nil {
+			return nil, err
+		}
+		networkInterfaces = append(networkInterfaces, intf)
+		return networkInterfaces, nil
+	}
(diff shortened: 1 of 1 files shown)


dcbw:
/bugzilla refresh
/lgtm
Leaving approve to somebody who is more familiar with Azure

kyrtapz:
Although I don't know much about azure security options I am in favor of **Option 2**, based on the following:

> By default, the first network interface attached to a VM is the primary network interface. All other network interfaces in the VM are secondary network interfaces.
https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-network-interface-vm

It's maybe worth adding a log message if it is not possible to get the primary interface.

@squeed any thoughts?

squeed:
I'm in favor of fail-closed: that is to say, if we're not able to determine the primary (and there is more than one), then we should not guess. We can report (ideally via a Status field) that we don't have the requisite cloud permissions (btw, should we fix our CloudCredentialsRequest somehow?), and thus cannot reconcile properly

andreaskaris:
Not familiar with Azure, so this is a very unqualified comment ... but in the case of > 1 interface, can we somehow iterate over the interface's network configuration and see if Azure provides a default route for that network? If only one of the attached interfaces has a default route, select that one, otherwise fail?

PR #31, opened 2022-04-06
Add instructions for how to run image in cloud

andreaskaris:

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,11 +193,44 @@ cloud.network.openshift.io/egress-ipconfig: [{"interface": "$IFNAME/$IFID", "ifa
 
 # How to hack/debug
 
-When debugging this component make sure to do the following:
+When you want to debug this component, use one of the following methods.
 
-1. Run the `./hack/run-locally.sh`, but make sure you have a `KUBECONFIG`
-   initialized in your environment variables. If running against an OpenShift
-   cluster: the credentials will be copied locally to your computer. That being
-   said, you probably won't be able to talk to the cloud API endpoints from your
-   local development environment, so you might need to comment out code which is
-   non-essential to your testing.
+## Run locally
+
+If you want to run the operator directly on your computer, you can
+run the `./hack/run-locally.sh`, but make sure you have a `KUBECONFIG`
+initialized in your environment variables. If running against an OpenShift
+cluster: the credentials will be copied locally to your computer. That being
+said, you probably won't be able to talk to the cloud API endpoints from your
+local development environment, so you might need to comment out code which is
+non-essential to your testing.
+
+## Patch the operator to run from a new image
+
+If you have issues talking directly to the cloud API endpoints or if you want
+to test how the operator runs inside the actual container, use this method.
+
+Set up your own publicly accessible repository. For example, use a quay repository:
+~~~
+export CNCC_REPOSITORY="quay.io/akaris/cloud-network-config-controller"
+~~~
+
+If you prefer using docker instead of podman, run:
+~~~
+export CONTAINER_ENGINE="docker"
+~~~
+
+Run:
+~~~
+hack/run_in_cloud.sh
+~~~
+
+This will then build the container from the Dockerfile, tag the image as `${CNCC_REPOSITORY}:${UUID}`
+where `${UUID}` is a unique ID, push the image to the registry and patch the clusterversion operator
+and the network-operator so that this new image will be used.
+
+In order to rollback, run `oc edit clusterversion version` and remove the configuration in
+`/spec/overrides`. The clusterversion operator should then bring the cluster back to its
+original state.
+
+> Note: If the cloud-network-config-controller pod runs into authentication problems after using this method, it may be necessary to delete the pod manually to "kick" it.
--- a/hack/run_in_cloud.sh
+++ b/hack/run_in_cloud.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+set -eux
+
+DIR=$(dirname "${BASH_SOURCE[0]}")
+
+CNCC_REPOSITORY=${CNCC_REPOSITORY:-""}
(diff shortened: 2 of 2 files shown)


hack/run_in_cloud.sh
```diff
@@ -0,0 +1,48 @@
...
+
+pushd $DIR/../
+	$BUILDAH_BUILD_CMD -t ${IMAGE}
+	$PODMAN_CMD push ${IMAGE}
+popd
+
+oc patch clusterversion version --type json -p '[{"op":"add","path":"/spec/overrides","value":[{"kind":"Deployment","group":"apps","name":"network-operator","namespace":"openshift-network-operator","unmanaged":true}]}]'
+oc scale -n openshift-network-operator deployment.apps/network-operator --replicas=0
+oc patch -n openshift-network-operator deployment.apps network-operator -p '{"spec":{"template":{"spec":{"containers":[{"name":"network-operator","env":[{"name":"CLOUD_NETWORK_CONFIG_CONTROLLER_IMAGE","value":"'${IMAGE}'"}]}]}}}}'
+oc describe  -n openshift-network-operator deployment.apps network-operator | grep CLOUD_NETWORK_CONFIG_CONTROLLER_IMAGE
```
jluhrsen, line 44: tiniest nit of all time. extra space between describe and -n. same happened below on lines 47 and 48 :)

hack/run_in_cloud.sh
```diff
@@ -0,0 +1,48 @@
...
+oc scale -n openshift-network-operator deployment.apps/network-operator --replicas=1
+sleep 30
```
jluhrsen, line 46: is there an 'oc wait' command to use instead of a sleep?

README.md
```diff
@@ -193,11 +193,31 @@ cloud.network.openshift.io/egress-ipconfig: [{"interface": "$IFNAME/$IFID", "ifa
...
 1. Run the `./hack/run-locally.sh`, but make sure you have a `KUBECONFIG`
    initialized in your environment variables. If running against an OpenShift
    cluster: the credentials will be copied locally to your computer. That being
    said, you probably won't be able to talk to the cloud API endpoints from your
    local development environment, so you might need to comment out code which is
-   non-essential to your testing.
\ No newline at end of file
+   non-essential to your testing.
+
+## Run from an new image
```
jluhrsen, line 207: can you add a blurb about what exactly this will actually do? ... patch deployment.apps network-operator with
custom image, etc.

hack/run_in_cloud.sh
andreaskaris: It's what happens when one copy/pastes from one's own ~/.bashrc hacks haha thanks!

hack/run_in_cloud.sh
andreaskaris: yep, done, thanks!

README.md
andreaskaris: done :-)

hack/run_in_cloud.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+# Note: Tested only with buildah / podman
```
danwinship, line 2: we don't need to support docker for our internal stuff. If you have only tested this with podman, just make it only work with podman

hack/run_in_cloud.sh
```diff
@@ -0,0 +1,75 @@
...
+
+set -eux
+
+DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
```
danwinship, line 6: Seems unnecessarily complicated. You build from `${DIR}/../` later, so you're assuming this is built from the source tree, but you know `hack/` isn't a symlink, so there's no need for the cd/pwd. (It shouldn't matter if anything above the checkout dir is a symlink.)

hack/run_in_cloud.sh
```diff
@@ -0,0 +1,75 @@
...
+
+if [ -z "${CNCC_REPOSITORY}" ]; then
+	echo "Please set environment variable CNCC_REPOSITORY and point it to your container repository"
+	exit 1
+fi
+
+CONTAINER_ENGINE="${CONTAINER_ENGINE:-podman}"
+BUILDAH_BIN="buildah"
+BUILDAH_BUILD_CMD="buildah bud"
```
danwinship, line 15: you can just use `podman build`, which doesn't require `buildah` to be installed

hack/run_in_cloud.sh
```diff
@@ -0,0 +1,75 @@
...
+IMAGE="${CNCC_REPOSITORY}:${UUID}"
+
+# retry logic for "Bad Gateway message" encountered a lot during testing
+push_container() {
+	f=$(mktemp)
+	for i in 1 2 3; do
+		$PODMAN_CMD push ${IMAGE} 2>&1 | tee $f
+		if [ ${PIPESTATUS[0]} -ne 0 ]; then
+			if grep -q "502 Bad Gateway" $f; then
+				echo "Got 502 Bad Gateway error, this was try: $i/3"
```
danwinship, line 43: you should not get these unless there is some sort of infrastructural error, in which case there is not reason to think that trying 3 times should help. Just push it, and if it fails, then bail out. The user can try again if they think there is reason to try again.

hack/run_in_cloud.sh
andreaskaris: done

hack/run_in_cloud.sh
andreaskaris: done

hack/run_in_cloud.sh
andreaskaris: done

andreaskaris:
Thanks for the review, Dan and Jamo!

andreaskaris:
/hold
Waiting for a cloud to test the new version on

PR #50, opened 2022-07-11
Bug 2104784: AWS: Fix race in IP address assignment code 

andreaskaris:

FYI I also talked to Alex about this fix here and about why it was originally done differently; Alex shot for idempotency, but did not consider this specific race condition here.

### DIFF
--- a/pkg/cloudprovider/aws.go
+++ b/pkg/cloudprovider/aws.go
@@ -14,6 +14,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/aws.go
```diff
@@ -71,20 +72,34 @@ func (a *AWS) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
...
 	keepIPs := []*string{}
+	newIPs := []*string{}
 	if utilnet.IsIPv6(ip) {
 		for _, assignedIPv6 := range networkInterface.Ipv6Addresses {
 			if assignedIP := net.ParseIP(*assignedIPv6.Ipv6Address); assignedIP != nil && assignedIP.Equal(ip) {
 				return AlreadyExistingIPError
 			}
 			keepIPs = append(keepIPs, assignedIPv6.Ipv6Address)
 		}
-		keepIPs = append(keepIPs, &addIP)
```
kyrtapz, line 81: What's the point of having `keepIPs` now? we don't use it to assign the new IP anyway

pkg/cloudprovider/aws.go
```diff
@@ -71,20 +72,34 @@ func (a *AWS) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
...
+		newIPs = append(newIPs, &addIP)
 		input := ec2.AssignIpv6AddressesInput{
 			NetworkInterfaceId: networkInterface.NetworkInterfaceId,
-			Ipv6Addresses:      keepIPs,
+			Ipv6Addresses:      newIPs,
```
kyrtapz, line 86: can we just create the array here and skip declaring `newIPs`?

pkg/cloudprovider/aws.go
```diff
@@ -71,20 +72,34 @@ func (a *AWS) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
...
+				keepIPsStr = fmt.Sprintf("%s %s", keepIPsStr, *ip)
+			}
+			klog.Warningf("error: %s, tried to assign the following %d IP addresses to the interface:%v. Already got %d IP addresses on the interface: %v",
+				err.Error(),
+				len(newIPs), newIPsStr,
+				len(keepIPs), keepIPsStr,
+			)
 			return err
 		}
 		return a.waitForCompletion(node, awsapi.StringValueSlice(keepIPs), false)
```
kyrtapz, line 105: I think this should change to `return a.waitForCompletion(node, awsapi.StringValueSlice(newIPs), false)`

pkg/cloudprovider/aws.go
andreaskaris: I renamed it to exsitingIPs and I added a comment above for why we need it. In go, we cant print the content of a slice of pointers, we'll get the only the pointers if we `fmt.Sprintf("%v", sliceOfPointers)`. So unless you know of something more elegant, we still must collect those IPs in a new []string{} slice

andreaskaris:
The failing EgressIP test is a flake in the test itself. I ran the test manually multiple times against this very PR and it passes without issues.
~~~
akaris@linux origin ((2ca9c0cc41...))]$ TEST_SUITE_START_TIME="1658754962" KUBE_TEST_REPO_LIST="" KUBE_TEST_REPO="quay.io/openshift/community-e2e-images" TEST_PROVIDER="{\"type\":\"aws\",\"ProjectID\":\"\",\"Region\":\"us-west-2\",\"Zone\":\"us-west-2b\",\"NumNodes\":3,\"MultiMaster\":true,\"MultiZone\":true,\"Zones\":[\"us-west-2b\",\"us-west-2c\"],\"ConfigFile\":\"\",\"Disconnected\":false,\"SingleReplicaTopology\":false,\"NetworkPlugin\":\"OpenShiftSDN\",\"HasIPv4\":true,\"HasIPv6\":false,\"HasSCTP\":false,\"IsProxied\":false,\"IsIBMROKS\":false}" TEST_JUNIT_DIR="" TEST_UPGRADE_OPTIONS="" ./openshift-tests run-test "[sig-network][Feature:EgressIP] [external-targets] EgressIPs can be assigned automatically [Skipped:Network/OVNKubernetes] [Serial] [Suite:openshift/conformance/serial]"
~~~

PR #51, opened 2022-07-25
Update OWNERS

trozet:
Make Patryk an approver.

Signed-off-by: Tim Rozet <trozet@redhat.com>


trozet:
@kyrtapz @andreaskaris PTAL

PR #53, opened 2022-08-17
Bug 2118563: Adjust OpenStack port capacity to default max_allowed_address_pair

andreaskaris:
The default max_allowed_address_pair is 10. That value cannot be
retrieved via the API so we must make a best guess here and set this to
the default.

Cause: 
The CloudNetworkConfigController (CNCC) used a hardcoded capacity of 64 which is above OpenStack's default max_allowed_address_pairs count of 10.

Consequence: 
Capacity was misreported and the EgressIP controller created CloudPrivateIPConfig objects beyond OpenStack's capacity. The CloudPrivateIPConfig objects would error out with "The number of allowed address pair exceeds the maximum 10"

Fix: 
The default max_allowed_address_pair for RHOSP is 10. Unfortunately, this value cannot be
retrieved via any API requests. It is solely exposed through neutron's configuration files. As a heuristic, the CNCC now assumes that all OSP environments set this to 10. As a consequence, any OSP environment that shall be used together with Red Hat OpenShift Container Platform and the EgressIP feature must have max_allowed_address_pairs set to 10 or above in neutron's configuration.

Result: 
Port capacity is now capped at 10 minus the number of allowed_address_pairs.

### DIFF
--- a/pkg/cloudprovider/openstack.go
+++ b/pkg/cloudprovider/openstack.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"math"
 	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/openstack.go
```diff
@@ -495,37 +492,14 @@ func (o *OpenStack) getNeutronPortNodeEgressIPConfiguration(p neutronports.Port)
...
-		if utilnet.IsIPv4(net.ParseIP(ip.IPAddress)) {
-			ipv4UsedIPs[ip.IPAddress] = struct{}{}
-		} else {
-			ipv6UsedIPs[ip.IPAddress] = struct{}{}
-		}
-	}
-
-	return len(ipv4UsedIPs), len(ipv6UsedIPs)
+// countAllowedAddressPairsOnPort returns the number of allowed_address_pairs on the port.
+func (o *OpenStack) countAllowedAddressPairsOnPort(p neutronports.Port) int {
```
kyrtapz, line 501: NIT: instead of having that function can we just use `len(p.AllowedAddressPairs)` instead?

flavio-fernandes: +1

flavio-fernandes (commented):
I agree with @kyrtapz on whether we really should add a function instead of just doing a len.
Other this looks good to me.

flavio-fernandes (commented):
/lgtm

PR #54, opened 2022-08-17
Jira OCPBUGS-208: OpenStack: Avoid concurrent port updates for attach/detach operations

andreaskaris:
Due to issue https://bugzilla.redhat.com/show_bug.cgi?id=2119199 in
neutron, we cannot rely on the revision_number checks that OSP provides.
Instead, we must implement our own per-port locking mechanism so that
concurrent updates become impossible.

This follows the same approach as the Azure fix in commit
https://github.com/openshift/cloud-network-config-controller/commit/3c2cb5063d7a1a9d60e48c6a48b4a42eab5da688.

Reported-at: https://issues.redhat.com/browse/OCPBUGS-208

### DIFF
--- a/pkg/cloudprovider/openstack.go
+++ b/pkg/cloudprovider/openstack.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/google/uuid"
 	"github.com/gophercloud/gophercloud"
@@ -52,8 +53,10 @@ const (
 // to the OpenStack API
 type OpenStack struct {
 	CloudProvider
-	novaClient    *gophercloud.ServiceClient
-	neutronClient *gophercloud.ServiceClient
+	novaClient       *gophercloud.ServiceClient
+	neutronClient    *gophercloud.ServiceClient
+	portLockMapMutex sync.Mutex
+	portLockMap      map[string]*sync.Mutex
 }
 
 // initCredentials initializes the cloud API credentials by reading the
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/openstack.go
```diff
@@ -634,6 +637,12 @@ func (o *OpenStack) getNeutronPortWithIPAddressAndMachineID(s neutronsubnets.Sub
 
 // allowIPAddressOnNeutronPort adds the specified IP address to the port's allowed_address_pairs.
 func (o *OpenStack) allowIPAddressOnNeutronPort(portID string, ip net.IP) error {
+	// Needed due to neutron bug  https://bugzilla.redhat.com/show_bug.cgi?id=2119199.
+	klog.Info("Getting port lock for portID %s", portID)
```
pperiyasamy, line 641: nit: can we also log ip address ?

pkg/cloudprovider/openstack.go
```diff
@@ -682,6 +691,12 @@ func (o *OpenStack) allowIPAddressOnNeutronPort(portID string, ip net.IP) error
 
 // unallowIPAddressOnNeutronPort removes the specified IP address from the port's allowed_address_pairs.
 func (o *OpenStack) unallowIPAddressOnNeutronPort(portID string, ip net.IP) error {
+	// Needed due to neutron bug  https://bugzilla.redhat.com/show_bug.cgi?id=2119199.
+	klog.Info("Getting port lock for portID %s", portID)
```
pperiyasamy, line 695: same here.

pkg/cloudprovider/openstack.go
```diff
@@ -682,6 +691,12 @@ func (o *OpenStack) allowIPAddressOnNeutronPort(portID string, ip net.IP) error
...
+	portLock := o.getLockForPort(portID)
+	portLock.Lock()
+	defer portLock.Unlock()
+
 	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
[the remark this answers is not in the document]
andreaskaris: neutron's built-in revision number mechanism does work when updates do not happen in super quick succession (I don't know what delays we're talking about, but we're somewhere in the milliseconds).
The retryonconflict and revision number mechanism shall also protect against a 3rd entity updating the port, *unless* we are talking about very short delays where BZ2119199. is an issue. So I'd prefer to leave this in here and just add the portlock mechanism on top of this until the neutron bug is resolved, if ever. And then the portlock should not be necessary [tbh, I doubt though that neutron will ever fix that bug]

PR #60, opened 2022-09-23
OCPBUGS-1629: Add custom resolver for setting endpoints

bertinatto:
This patch adds a custom resolver to only set a custom endpoint on EC2 service. This should prevent overwriting the STS endpoint, which causes the service to respond with 400 errors.

I'm not familiar with the openshift-cloud-network-config-controller code and can't test the changes, so please review carefully.

We are working on something similar for the EBS CSI Driver.

### DIFF
--- a/pkg/cloudprovider/aws.go
+++ b/pkg/cloudprovider/aws.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	awsapi "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


kyrtapz:
Thanks for the fix!

We need to adapt it a bit so we actually change the sessions resolver:
```
diff --git a/pkg/cloudprovider/aws.go b/pkg/cloudprovider/aws.go
index 4c508bb..c365ff1 100644
--- a/pkg/cloudprovider/aws.go
+++ b/pkg/cloudprovider/aws.go
@@ -29,11 +29,12 @@ type AWS struct {
 }
 
 func (a *AWS) initCredentials() error {
+       awsCfg := awsapi.NewConfig().WithRegion(a.cfg.Region)
        sessionOpts := session.Options{
                SharedConfigState: session.SharedConfigEnable,
                SharedConfigFiles: []string{filepath.Join(a.cfg.CredentialDir, "credentials")},
+               Config:            *awsCfg,
        }
-       c := awsapi.NewConfig().WithRegion(a.cfg.Region)
        if a.cfg.APIOverride != "" {
                customResolver := func(service, region string, optFns ...func(*endpoints.Options)) (endpoints.ResolvedEndpoint, error) {
                        if service == endpoints.Ec2ServiceID {
@@ -42,9 +43,10 @@ func (a *AWS) initCredentials() error {
                                        SigningRegion: a.cfg.Region,
                                }, nil
                        }
-                       return endpoints.DefaultResolver().EndpointFor(service, region, optFns...)
+                       ep, err := endpoints.DefaultResolver().EndpointFor(service, region, optFns...)
+                       return ep, err
                }
-               c.EndpointResolver = endpoints.ResolverFunc(customResolver)
+               sessionOpts.Config.EndpointResolver = endpoints.ResolverFunc(customResolver)
        }
        if a.cfg.AWSCAOverride != "" {
                var err error
@@ -58,8 +60,7 @@ func (a *AWS) initCredentials() error {
        if err != nil {
                return fmt.Errorf("could not initialize AWS session: %w", err)
        }
-
-       a.client = ec2.New(mySession, c)
+       a.client = ec2.New(mySession)
        return nil
 }
```

PR #61, opened 2022-09-26
OCPBUGS-1629: Add resolver to handle custom endpoints

kyrtapz:
Based on https://github.com/openshift/cloud-network-config-controller/pull/60

CNO sets `platform-api-url param`(`APIOverride`) to ec2 service endpoint https://github.com/openshift/cluster-network-operator/blob/e1d349aa0f2c1282bfb5ab843efe1f0501b2a9e2/pkg/network/cloud_network.go#L49
We should only use that endpoint for EC2 service.

This patch adds a custom resolver to only set a custom endpoint on EC2 service.

/assign @andreaskaris 

(cherry picked from commit 489c81d07e4bd5fbe93fd400cb27def4ab0ef70d)
Signed-off-by: Patryk Diak <pdiak@redhat.com>

### DIFF
--- a/pkg/cloudprovider/aws.go
+++ b/pkg/cloudprovider/aws.go
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	awsapi "github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


andreaskaris (changes requested):
nit for the commit message, can we change that to something along these lines:

AWS: Use platform-api-url override only for EC2 service endpoint

CNO sets platform-api-url param (APIOverride) to overwrite the EC2
service endpoint. Prior to this patch, the AWS cloud provider
unconditionally overwrote the endpoint URL for all endpoints. This 
caused issues with other service endpoints such as STS. Add a custom
resolver to only set the custom endpoint for EC2 service endpoint

kyrtapz:
Failures not related to CNCC changes:
```
{  2 events happened too frequently
event happened 21 times, something is wrong: ns/openshift-etcd-operator deployment/etcd-operator - reason/UnhealthyEtcdMember unhealthy members: ip-10-0-135-119.ec2.internal
event happened 22 times, something is wrong: ns/openshift-etcd-operator deployment/etcd-operator - reason/UnhealthyEtcdMember unhealthy members: ip-10-0-156-120.ec2.internal}
```
/override ci/prow/e2e-aws-serial

PR #68, opened 2022-10-12
OCPBUGS-1417: AWS: build temp credentials file from AWS key and secret

andreaskaris:
If file credentials cannot be found then build a temp file from aws_access_key_id and aws_secret_access_key.
Other operators build AWS credentials from aws_access_key_id and aws_secret_access_key without the need to have an explicit credentials file. This is not a problem for environments where the cloud-credential-operator injects the correct data. However, in manual credentials mode, users seem to expect to only provide the key_id and access_key.

Reported-at: https://issues.redhat.com/browse/OCPBUGS-1417

### DIFF
--- a/pkg/cloudprovider/aws.go
+++ b/pkg/cloudprovider/aws.go
@@ -1,6 +1,7 @@
 package cloudprovider
 
 import (
+	"bytes"
 	"fmt"
 	"net"
 	"os"
@@ -12,6 +13,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/aws.go
```diff
@@ -378,3 +385,53 @@ func getInstanceIdFromProviderId(providerId string) (string, error) {
...
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to read file %s", fAccessKeyID)
+	}
+	secretAccessKey, err := os.ReadFile(fSecretAccessKey)
+	if err != nil {
+		return "", errors.Wrapf(err, "failed to read file %s", fSecretAccessKey)
+	}
+	data := newConfigForStaticCreds(string(accessKeyID), string(secretAccessKey))
+
+	f, err := ioutil.TempFile("", "aws-shared-credentials")
```
kyrtapz, line 418: ioutil.TempFile is deprecated https://pkg.go.dev/io/ioutil#TempFile, can we use `os.CreateTemp` instead?

pkg/cloudprovider/aws.go
```diff
@@ -378,3 +385,53 @@ func getInstanceIdFromProviderId(providerId string) (string, error) {
 
 	return instanceId, nil
 }
+
+// sharedCredentialsFileFromDirectory returns a location (path) to the shared credentials file that was created
+// using the data in the provided directory.
+// This is heavily inspired and partially copied verbatim from openshift/cluster-ingress-operator, file
```
kyrtapz, line 391: Can we provide a link to the repo?

andreaskaris: done

pkg/cloudprovider/aws.go
andreaskaris: thanks done

PR #95, opened 2023-01-17
Add MovePrivateIP and its OpenStack implementation

dulek:
Up until now when an IP is failed over from one node to another, `CloudPrivateIPConfigController` will call `ReleasePrivateIP` with the node being failed over and `AssignPrivateIP` with the new node.

This is far from ideal for OpenStack platform. In this case in addition to adding the IP to the `allowedAddressPairs` on the node's port, we also create a port in Neutron to act as a reservation of that IP, so that it won't be taken by another node being created. With traditional model on every failover the reservation port would get removed, making a race condition "stealing" the IP possible. It's also problematic if user assigned a Neutron floating IP to that reservation port to have a predictible SNAT address for EgressIP traffic going outside the cluster. On failover such association would be lost and manual intervention would be needed to assign FIP to new reservation port.

This commit introduces an optional `CloudProviderWithMoveIntf` interface that has `MovePrivateIP` operation to be called on IP fail over. This operation is supposed to move the IP assignment from one node to another, but doesn't require deleting cloud resources if not necessary. The required logic is added to `CloudPrivateIPConfigController` to call `MovePrivateIP` on IP fail over if plugin implements the `CloudProviderWithMoveIntf` interface.

An implementation of `MovePrivateIP` is added to the OpenStack plugin, that will remove the IP from old node's `allowedAddressPairs`, add the IP to the new node's `allowedAddressPairs`, but won't touch the Neutron port that serves as a reservation of that IP.

### DIFF
--- a/pkg/cloudprovider/openstack.go
+++ b/pkg/cloudprovider/openstack.go
@@ -57,6 +57,7 @@ const (
 // to the OpenStack API
 type OpenStack struct {
 	CloudProvider
+	CloudProviderWithMoveIntf
 	novaClient       *gophercloud.ServiceClient
 	neutronClient    *gophercloud.ServiceClient
 	portLockMapMutex sync.Mutex
@@ -162,6 +163,77 @@ func (o *OpenStack) initCredentials() error {
 	return nil
 }
 
+// findAssignSubnetAndPort attempts to identify a subnet and server port for an AssignPrivateIP operation on a given node.
+func (o *OpenStack) findAssignSubnetAndPort(ip net.IP, node *corev1.Node) (*neutronsubnets.Subnet, *neutronports.Port, error) {
+	if node == nil {
+		return nil, nil, fmt.Errorf("invalid nil pointer provided for node when trying to assign private IP %s", ip.String())
+	}
+	// List all ports that are attached to this server.
+	serverID, err := getNovaServerIDFromProviderID(node.Spec.ProviderID)
+	if err != nil {
+		return nil, nil, err
+	}
+	serverPorts, err := o.listNovaServerPorts(serverID)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Loop over all ports that are attached to this nova instance and find the subnets
+	// that are attached to the port's network.
+	for _, serverPort := range serverPorts {
+		// If this IP address is already allowed on the port (speak: part of allowed_address_pairs),
+		// then return an AlreadyExistingIPError and skip all further steps.
+		if isIPAddressAllowedOnNeutronPort(serverPort, ip) {
+			// This is part of normal operation.
+			// Callers will likely ignore this and go on with their business logic and
+			// report success to the user.
+			return nil, nil, AlreadyExistingIPError
+		}
+
+		// Get all subnets that are attached to this port.
+		subnets, err := o.getNeutronSubnetsForNetwork(serverPort.NetworkID)
+		if err != nil {
+			klog.Warningf("Could not find subnet information for network %s, err: %q", serverPort.NetworkID, err)
+			continue
+		}
+		// 1) Loop over all subnets of the port and check if the IP address fits inside the subnet CIDR.
+		//   If the IP address is inside the subnet:
+		//   2) Return it as this is the subnet and port we were looking for.
+		// 3) Throw an error if the IP address does not fit in any of the attached network's subnets.
+		var matchingSubnet *neutronsubnets.Subnet
+		for _, s := range subnets {
+			// Because we're dealing with a pointer here for matchingSubnet:
+			// we must reassign s:= s or we'd overwrite the content that we point
+			// to.
+			s := s
+			// 1) Loop over all subnets and check if the IP address matches the subnet CIDR. If the IP
+			//    addresses matches multiple subnets on the same server port, then something is wrong
+			//    with this server's configuration and we should refuse to continue by throwing an error.
+			_, ipnet, err := net.ParseCIDR(s.CIDR)
+			if err != nil {
+				klog.Warningf("Could not parse subnet information %s for network %s, err: %q",
+					s.CIDR, serverPort.NetworkID, err)
+				continue
+			}
+			if !ipnet.Contains(ip) {
+				continue
+			}
+			if matchingSubnet != nil {
+				return nil, nil, fmt.Errorf("requested IP address %s for node %s and port %s matches 2 different subnets, %s and %s",
+					ip, node.Name, serverPort.ID, matchingSubnet.ID, s.ID)
+			}
+			matchingSubnet = &s
+		}
+		if matchingSubnet != nil {
+			// 2) Return the matching subnet and port.
+			return matchingSubnet, &serverPort, nil
+		}
+	}
+
+	// 3) The IP address does not fit in any of the attached networks' subnets.
+	return nil, nil, fmt.Errorf("could not assign IP address %s to node %s", ip, node.Name)
+}
+
 // AssignPrivateIP attempts to assigning the IP address provided to the VM
 // instance corresponding to the corev1.Node provided on the cloud the
 // cluster is deployed on.
@@ -183,8 +255,59 @@ func (o *OpenStack) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
 	if node == nil {
 		return fmt.Errorf("invalid nil pointer provided for node when trying to assign private IP %s", ip.String())
 	}
-	// List all ports that are attached to this server.
 	serverID, err := getNovaServerIDFromProviderID(node.Spec.ProviderID)
+	if err != nil {
+		return err
+	}
+
+	matchingSubnet, matchingPort, err := o.findAssignSubnetAndPort(ip, node)
+	if err != nil {
+		return err
+	}
+
+	// Reserve the IP address on the subnet by creating a new unattached neutron port.
+	unboundPort, err := o.reserveNeutronIPAddress(*matchingSubnet, ip, serverID)
+	if err != nil {
+		return err
+	}
+	// Then, add the IP address to the port's allowed_address_pairs.
+	//    TODO: use a more elegant retry mechanism.
+	if err = o.allowIPAddressOnNeutronPort(matchingPort.ID, ip); err != nil && !errors.Is(err, AlreadyExistingIPError) {
+		// Try to clean up the allocated port if adding the IP to allowed_address_pairs failed.
+		// Try this 10 times, but if this operation fails more than that, then user intervention is needed or
+		// the upper layer must call ReleasePrivateIP (because if the neutron port exists and holds
+		// a reservation, then the assign step will not continue after step 2).
+		var errRelease error
+		var releaseStatus string
+		for i := 0; i < 10; i++ {
+			errRelease = o.releaseNeutronIPAddress(*unboundPort, serverID)
+			// If the release operation was successful, then we are done.
+			if errRelease == nil {
+				releaseStatus = "Released neutron port reservation."
+				break
+			}
+			// Otherwise store the error message and retry.
+			releaseStatus = fmt.Sprintf("Could not release neutron port reservation after %d tries, err: %q", i+1, errRelease)
+		}
+		return fmt.Errorf("could not allow IP address %s on port %s, err: %q. %s", ip.String(), matchingPort.ID, err, releaseStatus)
+	}
+	// Return nil to indicate success if steps 2 and 3 passed.
+	return nil
+}
+
+// MovePrivateIP implements moving the IP from one node to another to serve cases like a failover.
+// What's different from calling ReleasePrivateIP followed by AssignPrivateIP is that the reservation
+// Neutron port will not get deleted - MovePrivateIP will only change the allowed_address_pairs on the node's
+// ports to remove IP from nodeToDel and add it to nodeToAdd and update the existing reservation port with the DeviceID
+// of nodeToAdd. Additionally, if reservation port is missing MovePrivateIP will attempt to recreate it (this is a
+// corner case and should not happen in normal operation).
+func (o *OpenStack) MovePrivateIP(ip net.IP, nodeToAdd, nodeToDel *corev1.Node) error {
+	if nodeToAdd == nil || nodeToDel == nil {
+		return fmt.Errorf("invalid nil pointer provided for node when trying to move IP %s", ip.String())
+	}
+
+	// List all ports that are attached to this server.
+	serverID, err := getNovaServerIDFromProviderID(nodeToDel.Spec.ProviderID)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
```diff
@@ -179,24 +179,98 @@ func (c *CloudPrivateIPConfigController) SyncHandler(key string) error {
 		return err
 	}
 
-	// At most one of nodeToAdd or nodeToDel will be set
-	nodeToAdd, nodeToDel := c.computeOp(cloudPrivateIPConfig)
+	// At most one of nodeNameToAdd or nodeNameToDel will be set
```
mandre, line 182: No longer true with your change. Please update the comments.

mandre (commented):
Mostly comments about comments :)
The code looks good to me. I'm not quite sure what causes [the unit tests failure](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_cloud-network-config-controller/95/pull-ci-openshift-cloud-network-config-controller-master-unit/1615389218903691264) but it doesn't look related to this patch as it failed in a similar fashion [for two other PRs recently](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-cloud-network-config-controller-master-unit).

pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
```diff
@@ -179,24 +179,98 @@ func (c *CloudPrivateIPConfigController) SyncHandler(key string) error {
...
+		}
+		if cloudPrivateIPConfig, err = c.updateCloudPrivateIPConfigStatus(cloudPrivateIPConfig, status); err != nil {
+			return fmt.Errorf("error updating CloudPrivateIPConfig: %q during move operation, err: %v", key, err)
+		}
 
-		node, err := c.nodesLister.Get(nodeToDel)
+		// This is a blocking call. If the IP is not assigned then don't treat
+		// it as an error.
+		var i interface{} = c.cloudProviderClient
+		withMover, ok := i.(cloudprovider.CloudProviderWithMoveIntf)
```
andreaskaris, line 219: nit: I find the below way easier to read, but that may be personal preference:
~~~
withMover, ok := c.cloudProviderClient.(cloudprovider.CloudProviderWithMoveIntf)
~~~

pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
```diff
@@ -179,24 +179,98 @@ func (c *CloudPrivateIPConfigController) SyncHandler(key string) error {
...
+		if !ok {
+			return fmt.Errorf("cannot convert driver to the interface with move abilities, this should never happen")
+		}
+		if moveErr := withMover.MovePrivateIP(ip, nodeToAdd, nodeToDel); moveErr != nil && !errors.Is(moveErr, cloudprovider.NonExistingIPError) {
```
andreaskaris, line 223: extreme nit (also for other lines): I like keeping lines at 120 chars max + a few if I really cannot avoid it (which may be the case here). I know that's not the case for the older code in the repo and we don't really have a standard, but it's easier to read IMO. Feel free to ignore this and keep as is

andreaskaris: Looking at the code, I think ReleasePrivateIP is the only function atm that can throw the NonExistingIPError (which is logical, in case of a Release, we have a means of reporting the issue, but we ignore it). I'm not sure if it makes sense for MovePrivateIP to throw such an error. Currently as far as I see it doesn't, and so we don't have to deal with it hear, either?

pkg/cloudprovider/openstack.go
```diff
@@ -237,47 +222,127 @@ func (o *OpenStack) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
...
+	}
+	serverPorts, err := o.listNovaServerPorts(serverID)
+	if err != nil {
+		return err
+	}
+
+	// Loop over all ports that are attached to this nova instance.
+	for _, serverPort := range serverPorts {
+		if isIPAddressAllowedOnNeutronPort(serverPort, ip) {
+			if err = o.unallowIPAddressOnNeutronPort(serverPort.ID, ip); err != nil {
```
andreaskaris, line 326: Does the comment in 332 refer to this line here? If so, I'd say we need to return the error, because we check if the IP is in the allowed_address_pairs, and only if that's the case, we try to remove it (and retry). So if this still fails, then something went wrong during the detach / update of allowed_address_pairs. Returning the error will make sure that we do not continue without removing the IP from the allowed_address_pair on the source port. This is also important because without this succeeding, findAssignSubnetAndPort might return AlreadyExistingIPError

pkg/cloudprovider/openstack.go
```diff
@@ -237,47 +222,127 @@ func (o *OpenStack) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
...
 		}
+		// 4) Return nil to indicate success if steps 2 and 3 passed.
+		return nil
 	}
 
 	// 5) The IP address does not fit in any of the attached networks' subnets.
 	return fmt.Errorf("could not assign IP address %s to node %s", ip, node.Name)
 }
 
+func (o *OpenStack) MovePrivateIP(ip net.IP, nodeToAdd, nodeToDel *corev1.Node) error {
```
andreaskaris, line 308: I think what's missing is that we must account for the crazy corner case where the reservation port is gone (perhaps an admin deleted it) and must be recreated by MovePrivateIP?

pkg/cloudprovider/openstack.go
```diff
@@ -237,47 +222,127 @@ func (o *OpenStack) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
 				continue
 			}
 			if matchingSubnet != nil {
-				return fmt.Errorf("requested IP address %s for node %s and port %s matches 2 different subnets, %s and %s",
+				return nil, nil, fmt.Errorf("requested IP address %s for node %s and port %s matches 2 different subnets, %s and %s",
 					ip, node.Name, serverPort.ID, matchingSubnet.ID, s.ID)
 			}
+
```
andreaskaris, line 228: nit: blank line needed?

pkg/cloudprovider/openstack.go
```diff
@@ -237,47 +222,127 @@ func (o *OpenStack) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
...
+	if err != nil {
+		return err
+	}
+
+	matchingSubnet, matchingPort, err := o.findAssignSubnetAndPort(ip, node)
+	if err != nil {
+		return err
+	}
+
+	if matchingSubnet != nil {
```
andreaskaris, line 273: findAssignSubetAndPort either returns nil,nil,error   or non-nil, non-nil, nil as far as I can see, so we can simplify this and remove the if and the return error in 5)?

andreaskaris (changes requested):
It'd be great to unit test `MovePrivateIP` in test function `TestOpenStackPlugin`

pkg/cloudprovider/openstack.go
andreaskaris, line 308: linter nit: openstack.go:303:1: exported method OpenStack.MovePrivateIP should have comment (...)

pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
dulek: I removed it and added a detailed comment to computeOp.

pkg/cloudprovider/openstack.go
dulek: Then we should probably take care of a crazy corner case when the reservation port exists in AssignPrivateIP too. I'll look into this.

pkg/cloudprovider/openstack.go
dulek: We do remove the error. The comment is about - what if for each of the ports `isIPAddressAllowedOnNeutronPort()` was false?

pkg/cloudprovider/openstack.go
```diff
@@ -237,47 +222,127 @@ func (o *OpenStack) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
...
-						releaseStatus = "Released neutron port reservation."
-						break
-					}
-					// Otherwise store the error message and retry.
-					releaseStatus = fmt.Sprintf("Could not release neutron port reservation after %d tries, err: %q", i+1, errRelease)
+			return matchingSubnet, &serverPort, nil
+		}
+	}
+
+	// 5) The IP address does not fit in any of the attached networks' subnets.
```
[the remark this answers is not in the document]
dulek: True, thanks!

pkg/cloudprovider/openstack.go
```diff
@@ -554,7 +617,38 @@ func (o *OpenStack) reserveNeutronIPAddress(s neutronsubnets.Subnet, ip net.IP,
...
+				return p, fmt.Errorf(
+					"bogus result from Neutron, multiple ports with IP %s in subnet %s", ip.String(), s.ID)
+			} else if len(ports) == 0 {
+				// Apparently it's not our port, return the original error.
+				return nil, err
+			}
+			// Otherwise we've got our port, let's return it.
+			return &ports[0], nil
+		}
+		// Regular error, just return it.
```
dulek, line 651: This is part I'll need to test - not sure about the error gophercloud returns in such case.

pkg/cloudprovider/openstack.go
```diff
@@ -554,7 +617,38 @@ func (o *OpenStack) reserveNeutronIPAddress(s neutronsubnets.Subnet, ip net.IP,
...
+		if _, ok := err.(*gophercloud.ErrDefault409); ok {
+			// If so, let's get the port and check if it's our reservation port. It's possible we've created it earlier.
+			// It's also possible user configured a taken range, we should erorr out in such case.
+			page, nestedErr := neutronports.List(o.neutronClient, neutronports.ListOpts{
+				NetworkID:   s.NetworkID,
+				FixedIPs:    []neutronports.FixedIPOpts{{SubnetID: s.ID, IPAddress: ip.String()}},
+				DeviceOwner: egressIPTag,
+				DeviceID:    generateDeviceID(serverID),
+			}).AllPages()
+			if nestedErr != nil {
```
dulek, line 632: BTW - what's the correct pattern to nest errors? This one smells bad. I guess a separate function might look better?

pkg/cloudprovider/openstack.go
andreaskaris: touché ;-)

pkg/cloudprovider/openstack.go
andreaskaris: you can wrap errors with %w:
https://rollbar.com/blog/golang-wrap-and-unwrap-error/#

There's a repo that implements this:
github.com/pkg/errors
With Wrap functions

You can join multiple errors with:
import (
apierrors "k8s.io/apimachinery/pkg/util/errors"
)
 apierrors.NewAggregate(errors)
https://pkg.go.dev/k8s.io/apimachinery/pkg/util/errors#NewAggregate

pkg/cloudprovider/openstack.go
dulek: I meant no harm, it's a legitimate corner case. ;)

pkg/cloudprovider/openstack.go
```diff
@@ -237,45 +219,126 @@ func (o *OpenStack) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
...
+		return fmt.Errorf("could not allow IP address %s on port %s, err: %q. %s", ip.String(), matchingPort.ID, err, releaseStatus)
+	}
+	// 4) Return nil to indicate success if steps 2 and 3 passed.
+	return nil
+}
+
+// MovePrivateIP implements moving the IP from one node to another to serve cases like a failover.
+// What's different from calling ReleasePrivateIP followed by AssignPrivateIP is that the reservation
+// Neutron port will not get deleted - MovePrivateIP will only change the allowed_address_pairs on the node's
+// ports to remove IP from nodeToDel and add it to nodeToAdd.
```
andreaskaris, line 301: Add something like: "Additionally, MovePrivateIP will recreate a missing reservation port (this is a corner case and should not happen in normal operation)"

pkg/cloudprovider/openstack.go
```diff
@@ -554,7 +617,38 @@ func (o *OpenStack) reserveNeutronIPAddress(s neutronsubnets.Subnet, ip net.IP,
 		Name:        fmt.Sprintf("egressip-%s", ip.String()),
 	}
 	p, err := neutronports.Create(o.neutronClient, opts).Extract()
+
 	if err != nil {
+		// Let's check if error suggests that port with that IP already exists in that subnet.
+		if _, ok := err.(*gophercloud.ErrDefault409); ok {
+			// If so, let's get the port and check if it's our reservation port. It's possible we've created it earlier.
+			// It's also possible user configured a taken range, we should erorr out in such case.
```
andreaskaris, line 625: erorr -> error

pkg/cloudprovider/openstack.go
```diff
@@ -237,45 +219,126 @@ func (o *OpenStack) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
...
 
-	// 5) The IP address does not fit in any of the attached networks' subnets.
-	return fmt.Errorf("could not assign IP address %s to node %s", ip, node.Name)
+	subnet, port, err := o.findAssignSubnetAndPort(ip, nodeToAdd)
+	if err != nil {
+		return err
+	}
+
+	// This call is to double-check if the reservation port exists, function
+	// will return okay if the port with that IP already exists in the subnet.
```
andreaskaris, line 332: ... will return okay if the port with that IP already exists in the subnet, otherwise it will create the missing reservation port.

pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
```diff
@@ -179,24 +179,96 @@ func (c *CloudPrivateIPConfigController) SyncHandler(key string) error {
 		return err
 	}
 
-	// At most one of nodeToAdd or nodeToDel will be set
-	nodeToAdd, nodeToDel := c.computeOp(cloudPrivateIPConfig)
+	nodeNameToAdd, nodeNameToDel := c.computeOp(cloudPrivateIPConfig)
```
kyrtapz, line 182: NIT: For future reference, it is generally better to do variable renaming as a separate commit. It makes reviewing actual changes a lot easier.
You don't need to change it in this PR.

pkg/cloudprovider/openstack.go
```diff
@@ -237,45 +219,126 @@ func (o *OpenStack) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
...
+	if err != nil {
+		return err
+	}
+
+	matchingSubnet, matchingPort, err := o.findAssignSubnetAndPort(ip, node)
+	if err != nil {
+		return err
+	}
+
+	// 2) Reserve the IP address on the subnet by creating a new unattached neutron port.
```
kyrtapz, line 268: Do the numbers in comments still make sense in `AssignPrivateIP` I don't think we need them with the introduction of `findAssignSubnetAndPort`

pkg/cloudprovider/openstack.go
```diff
@@ -554,7 +617,38 @@ func (o *OpenStack) reserveNeutronIPAddress(s neutronsubnets.Subnet, ip net.IP,
...
+			if len(ports) > 1 {
+				// This is unexpected, Neutron should not allow multiple ports with the same IP
+				// in one subnet
+				return p, fmt.Errorf(
+					"bogus result from Neutron, multiple ports with IP %s in subnet %s", ip.String(), s.ID)
+			} else if len(ports) == 0 {
+				// Apparently it's not our port, return the original error.
+				return nil, err
+			}
+			// Otherwise we've got our port, let's return it.
```
kyrtapz, line 648: can we add a info log saying we will reuse an existing port?

pkg/cloudprovider/openstack.go
dulek: I just changed the logic instead, I wasn't exactly wanting to wrap errors.

dulek:
I think I addressed all of the comments. I still have an issue with ovn-kubernetes, where CloudPrivateIPConfig is getting deleted after a successful move operation. I'm not sure if this isn't an ovn-kubernetes bug and I'll be digging that.

dulek:
I fixed ovn-kubernetes issue here: https://github.com/ovn-org/ovn-kubernetes/pull/3403. Please take a look.

dulek:
@kyrtapz, @andreaskaris: https://github.com/ovn-org/ovn-kubernetes/pull/3403 is merged, so I think we can reconsider this one.

PR #96, opened 2023-01-25
OpenStack: Fix race condition in TestGetNodeEgressIPConfiguration

andreaskaris:
Fix a race condition in TestGetNodeEgressIPConfiguration unit test due to unstable map order.

### DIFF
--- a/pkg/cloudprovider/openstack_test.go
+++ b/pkg/cloudprovider/openstack_test.go
@@ -738,9 +738,9 @@ func TestGetNodeEgressIPConfiguration(t *testing.T) {
 	}
 
 	tcs := map[string]struct {
-		node                       *corev1.Node
-		expectedNodeEgressIPConfig []NodeEgressIPConfiguration
-		errString                  string
+		node                        *corev1.Node
+		expectedNodeEgressIPConfigs [][]NodeEgressIPConfiguration
+		errString                   string
 	}{
 		"Invalid node": {
 			node: &corev1.Node{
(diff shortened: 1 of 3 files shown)


pkg/cloudprovider/openstack_test.go
```diff
@@ -409,8 +410,19 @@ func portListHandler(t *testing.T, w http.ResponseWriter, r *http.Request) {
 	deviceOwner := r.URL.Query().Get("device_owner")
 	networkID := r.URL.Query().Get("network_id")
 
+	// The "undefined" test case in TestGetNodeEgressIPConfiguration relies on stable sort order of the returned neutron
```
kyrtapz, line 413: Can you use `"Valid node 3 - undefined behavior, IPv4 InternalIP and IPv6 InternalIP on different networks"` instead of  `"undefined"`?

pkg/cloudprovider/openstack_test.go
```diff
@@ -409,8 +410,19 @@ func portListHandler(t *testing.T, w http.ResponseWriter, r *http.Request) {
...
+	// interfaces. Let's make our lives easier and let's suppose that neutron always returns ports in stable order for
```
kyrtapz, line 414: What is neutrons behavior in general? does it return ports in a stable order?
I am fine with the approach you took but an alternative would be to expand `expectedNodeEgressIPConfig` to allow for more than one return possibility.

pkg/cloudprovider/openstack_test.go
andreaskaris: done

pkg/cloudprovider/openstack_test.go
andreaskaris: ok I expanded expectedNodeEgressIPConfig instead  - it doesn't really make a difference to me

andreaskaris:
@dulek @kyrtapz Please take another look

PR #103, opened 2023-04-14
OCPBUGS-11187: sync CloudPrivateIpConfig when node is missing

jluhrsen:
in the case that an egressIP is assigned to a node and that node is deleted, the sync will not be able to Get() that node but we still need to release
that IP and be able assign that to another node
in the egressIP object. The machine must be deleted as well, although that should be what triggers
the node delete.

### DIFF
--- a/pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
+++ b/pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
@@ -264,7 +264,12 @@ func (c *CloudPrivateIPConfigController) SyncHandler(key string) error {
 		klog.Infof("CloudPrivateIPConfig: %q will be deleted from node: %q", key, nodeNameToDel)
 
 		node, err := c.nodesLister.Get(nodeNameToDel)
-		if err != nil {
+		// there is a case when the node was deleted and the ip still needs to be released. if the node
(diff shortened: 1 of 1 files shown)


pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
```diff
@@ -264,7 +264,11 @@ func (c *CloudPrivateIPConfigController) SyncHandler(key string) error {
 		klog.Infof("CloudPrivateIPConfig: %q will be deleted from node: %q", key, nodeNameToDel)
 
 		node, err := c.nodesLister.Get(nodeNameToDel)
-		if err != nil {
+		// there is a case when the node was deleted and the ip still needs to be released. if the node
+		// doesn't exist, nodesLister.Get() will still return and error but we don't want to return with it
+		if node == nil {
```
kyrtapz, line 269: This comparison is not correct.
If err is set it is very likely that node will be set to nil, it would make us always enter the first if.
Instead please add an additional check under `if err != nil` that checks the error itself:
```
if err != nil {
    if apierrors.IsNotFound(err) {
        klog.Infof("Node: %s no longer exists. Will still need to unassign CloudPrivateIPConfig: %q", nodeNameToDel, key)
    } else {
        return err
    }
}

```

PR #112, opened 2023-05-18
[release-4.12] OCPBUGS-13802: sync CloudPrivateIpConfig when node is missing

jluhrsen:
in the case that an egressIP is assigned to a node and that node is deleted, the sync will not be able to Get() that node but we still need to release
that IP and be able assign that to another node
in the egressIP object. The machine must be deleted as well, although that should be what triggers
the node delete.


jluhrsen:
I have tested this locally and verified it works as expected.

kyrtapz:
@jluhrsen unit tests are failing :( 
When there are conflicts during backports it is good to mention them in the commit/pr description. It allows reviewers to focus on those areas. 

/retitle [release-4.12] OCPBUGS-13802: sync CloudPrivateIpConfig when node is missing

jluhrsen:
retest

PR #115, opened 2023-07-05
OCPBUGS-15805: Azure: Handle already existing IP configurations

kyrtapz:
If a secondary IP is already configured CNCC should return AlreadyExistingIPError instead of trying to add the same IP again.


flavio-fernandes (commented):
/lgtm

PR #119, opened 2023-07-20
[release-4.10] OCPBUGS-16626: pull project name from subnet uri

jluhrsen:
In the case of GCP XPN the subnet project is not the same as the project that is determined from the providerID of the node. That is because the network resources are from a shared-vpc and come from a different "shared" project.


jluhrsen:
> /override ci/prow/e2e-aws-ovn /override ci/prow/e2e-aws-sdn-serial

they are not even listed as required here, but tide is confused too I guess. Maybe I will have to do this all over from scratch and a new branch name.

PR #120, opened 2023-08-29
OCPBUGS-17151: AWS: Skipping Unusable Network Interfaces

flavio-fernandes:
When obtaining the network interfaces for a given node, it is possible that the returned slice contains entries with invalid subnet IDs.

### DIFF
--- a/pkg/cloudprovider/aws.go
+++ b/pkg/cloudprovider/aws.go
@@ -193,7 +193,12 @@ func (a *AWS) GetNodeEgressIPConfiguration(node *corev1.Node, cloudPrivateIPConf
 	if err != nil {
 		return nil, err
 	}
+	// For now, we only consider the first interface for EgressIP
 	networkInterface := networkInterfaces[0]
+	if networkInterface.NetworkInterfaceId == nil || *networkInterface.NetworkInterfaceId == "" ||
+		networkInterface.SubnetId == nil || *networkInterface.SubnetId == "" {
+		return nil, fmt.Errorf("error retrieving network interface with valid ids: %v", networkInterfaces)
+	}
 	config := &NodeEgressIPConfiguration{
 		Interface: *networkInterface.NetworkInterfaceId,
 	}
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/aws.go
```diff
@@ -263,7 +279,11 @@ func (a *AWS) getSubnet(networkInterface *ec2.InstanceNetworkInterface) (*net.IP
 		return nil, nil, fmt.Errorf("error: cannot list ec2 subnets, err: %v", err)
 	}
 	if len(describeOutput.Subnets) > 1 {
-		return nil, nil, fmt.Errorf("error: multiple subnets found for the subnet ID: %s", *networkInterface.SubnetId)
+		subnetId := "<nil>"
+		if networkInterface.SubnetId != nil {
```
flavio-fernandes, line 283: Note: with change in line 196, this is not possible unless getSubnet() starts to be called from another location. Maybe we don't need it here anymore?

pkg/cloudprovider/aws.go
```diff
@@ -193,7 +193,23 @@ func (a *AWS) GetNodeEgressIPConfiguration(node *corev1.Node, cloudPrivateIPConf
 	if err != nil {
 		return nil, err
 	}
-	networkInterface := networkInterfaces[0]
+	// Locate the first network interface with a valid interface and subnet for given node
+	var networkInterface *ec2.InstanceNetworkInterface
```
kyrtapz, line 197: Prior to your changes we only considered the first interface, we do something very similar on other clouds where we do not allow (yet) configuring egress IPs on secondary interfaces.
What is the reason for considering all interfaces?

flavio-fernandes: Good question. My reasoning was along the lines of making the code a bit mode future proof.
If the implementation is best being strict in only expecting the first interface to be supported,
we can do it that way for sure. Let me rewrite it to behave that way.

pkg/cloudprovider/aws.go
kyrtapz: Have you checked whether `len(describeOutput.Subnets) > 1` can be true when `SubnetId` is set?

pkg/cloudprovider/aws.go
```diff
@@ -193,7 +193,13 @@ func (a *AWS) GetNodeEgressIPConfiguration(node *corev1.Node, cloudPrivateIPConf
 	if err != nil {
 		return nil, err
 	}
+	// For now, we only consider the first interface for EgressIP
 	networkInterface := networkInterfaces[0]
+	if networkInterface == nil ||
```
kyrtapz, line 198: first interface is already checked for nil in:
```
func (a *AWS) getNetworkInterfaces(instance *ec2.Instance) ([]*ec2.InstanceNetworkInterface, error) {
	if len(instance.NetworkInterfaces) == 0 {
		return nil, NoNetworkInterfaceError
	}
	if instance.NetworkInterfaces[0] == nil {
		return nil, NoNetworkInterfaceError
	}
	return instance.NetworkInterfaces, nil
}
```

pkg/cloudprovider/aws.go
kyrtapz: Looking at the implementation it shouldn't return more than one subnet if there is one non-nil subnet provided but that check is not really hurting us so I am ok either way.

pkg/cloudprovider/aws.go
flavio-fernandes: oh, good point. Let me remove the redundant check.

PR #121, opened 2023-08-30
OCPBUGS-5491: Azure: skip backend pool if attached to an outbound rule

ricky-rav:
When configuring egress IP on Azure, omit the backend address pool if it's attached to outbound rule, since attaching a secondary IP is not allowed in such cases on Azure, resulting in an `OutboundRuleCannotBeUsedWithBackendAddressPoolThatIsReferencedBySecondaryIpConfigs` error.

This affects the following setups:
- azure private clusters
- ARO private clusters
- ARO public custers

**WARNING: as a result, the egress IP will have no outbound connectivity except to the infrastructure subnet. Further investigation in the form of an RFE is needed to let the egress IP communicate to the outside.**

### DIFF
--- a/pkg/cloudprovider/azure.go
+++ b/pkg/cloudprovider/azure.go
@@ -45,6 +45,7 @@ type Azure struct {
 	vmClient                     compute.VirtualMachinesClient
 	virtualNetworkClient         network.VirtualNetworksClient
 	networkClient                network.InterfacesClient
+	backendAddressPoolClient     network.LoadBalancerBackendAddressPoolsClient
 	nodeMapLock                  sync.Mutex
 	nodeLockMap                  map[string]*sync.Mutex
 	azureWorkloadIdentityEnabled bool
@@ -125,6 +126,12 @@ func (a *Azure) initCredentials() error {
 	a.virtualNetworkClient = network.NewVirtualNetworksClientWithBaseURI(a.env.ResourceManagerEndpoint, subscriptionID)
 	a.virtualNetworkClient.Authorizer = authorizer
 	_ = a.virtualNetworkClient.AddToUserAgent(UserAgent)
+
+	a.backendAddressPoolClient = network.NewLoadBalancerBackendAddressPoolsClientWithBaseURI(
+		a.env.ResourceManagerEndpoint, subscriptionID)
+	a.backendAddressPoolClient.Authorizer = authorizer
+	_ = a.backendAddressPoolClient.AddToUserAgent(UserAgent)
+
 	return nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/azure.go
```diff
@@ -152,14 +159,59 @@ func (a *Azure) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
 	ipConfigurations := *networkInterface.IPConfigurations
 	name := fmt.Sprintf("%s_%s", node.Name, ipc)
 	untrue := false
+
+	// In some azure setups (Azure private, public ARO, private ARO) outbound connectivity is achieved through
```
flavio-fernandes, line 163: NIT:
- "Azure" should be consistently capitalized.
- Add a comma before "outbound": `In public azure clusters, outbound connectivity is achieved through`

pkg/cloudprovider/azure.go
```diff
@@ -152,14 +159,59 @@ func (a *Azure) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
...
+				// attached to it
+				realPool, err := a.getBackendAddressPool(*pool.ID)
+				if err != nil {
+					return fmt.Errorf("error looking up backend address pool %s with ID %s: %v", *pool.Name, *pool.ID, err)
+				}
+				if realPool.BackendAddressPoolPropertiesFormat != nil {
+					if realPool.BackendAddressPoolPropertiesFormat.OutboundRule != nil {
+						loadBalancerBackendAddressPoolsArgument = nil
+						attachedOutboundRule = realPool.BackendAddressPoolPropertiesFormat.OutboundRule
+						break
```
flavio-fernandes, line 191: This will break out of the inner for loop (line 175). Should it also break us out of the outer for loop (line 174)? If so, maybe we could add a check for `loadBalancerBackendAddressPoolsArgument` in line 199?

or use a label? Example:
```
NeedleSearch:
	for _, stack := range haystack {
		for _, straw := range stack {
			if straw == 10 { // Found the needle!
				fmt.Println("Found the needle!")
				break NeedleSearch
			}
		}
	}
```

pkg/cloudprovider/azure.go
```diff
@@ -380,6 +432,47 @@ func (a *Azure) getNetworkInterfaces(instance *compute.VirtualMachine) ([]networ
 	return networkInterfaces, nil
 }
 
+func splitObjectID(azureResourceID string) (string, string, string) {
```
flavio-fernandes, line 435: NIT this is just a suggestion:
```
func splitObjectID(azureResourceID string) (resourceGroupName, loadBalancerName, backendAddressPoolName string) {
```
and then remove line 443 and 460 is just `return` because we know what these returning strings are :)

flavio-fernandes (commented):
Very nice @ricky-rav !  I just have some nits / suggestions I would like you to consider.
It kinda sucks that we do not have `azure_test.go`... I will not force you do create one, but do you think we should have one added for exercising the functions you are adding here?

pkg/cloudprovider/azure.go
ricky-rav: sure!

pkg/cloudprovider/azure.go
ricky-rav: sure!

trozet (approved):
/approve

PR #132, opened 2024-02-28
OCPBUGS-29915: GCP: Explicitly set UniverseDomain to the default universe domain

kyrtapz:
If the UniverseDomain is not set, the client will try to retrieve it from the metadata server: https://github.com/openshift/cloud-network-config-controller/blob/dc255162b1442a1b85aa0b2ab37ed63245857476/vendor/golang.org/x/oauth2/google/default.go#L77-L113
This won't work in OpenShift because the CNCC pod cannot access the metadata service IP address (we block the access to 169.254.169.254 from cluster-networked pods). 
Set the UniverseDomain to the default value explicitly.
Another option would be to make cncc a host networked pod but the result would be the same as in OCP we hit this:
```
// If the GCE metadata server returns a 404 error, the default value is
// returned. If the GCE metadata server returns an error other than 404, the
// error is returned.


curl -s -o /dev/null -w "%{http_code}"  -H "Metadata-Flavor: Google" http://169.254.169.254/computeMetadata/v1/universe/universe_domain
404
```

### DIFF
--- a/pkg/cloudprovider/gcp.go
+++ b/pkg/cloudprovider/gcp.go
@@ -20,6 +20,9 @@ const (
 	// already in use by default and subtract from 100. See:
 	// https://cloud.google.com/vpc/docs/quota#per_instance .
 	defaultGCPPrivateIPCapacity = 100
+	// default universe domain
+	// https://github.com/openshift/cloud-network-config-controller/blob/dc255162b1442a1b85aa0b2ab37ed63245857476/vendor/golang.org/x/oauth2/google/default.go#L25
+	defaultUniverseDomain = "googleapis.com"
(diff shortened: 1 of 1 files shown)


martinkennelly:
Seems fine. I dont think that address will change anytime soon.
LGTM

kyrtapz:
/hold 
lets wait for GCP to pass

kyrtapz:
https://redhat-internal.slack.com/archives/CEKNRGF25/p1709812331857849 :(

kyrtapz:
GCP is passing
/hold cancel

martinkennelly (approved):

PR #133, opened 2024-03-08
OCPBUGS-27422: Avoid nil pointer panic while assigning private IP on Azure

arghosh93:
If any previously assigned IP present in IP configurations of the NIC is removed from public load balancer backend pool then cloud-network-config-controller starts to panic due to a nil pointer.

This PR checks whether load balancer backend pull is nil or not before proceeding further.

### DIFF
--- a/pkg/cloudprovider/azure.go
+++ b/pkg/cloudprovider/azure.go
@@ -170,9 +170,9 @@ func (a *Azure) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
 	// UserDefinedRouting, which doesn't impose such constraints on secondary IPs.
 	loadBalancerBackendAddressPoolsArgument := (*networkInterface.IPConfigurations)[0].LoadBalancerBackendAddressPools
 	var attachedOutboundRule *network.SubResource
-	if (*networkInterface.IPConfigurations)[0].LoadBalancerBackendAddressPools != nil {
-	OuterLoop:
-		for _, ipconfig := range *networkInterface.IPConfigurations {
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/azure.go
```diff
@@ -170,9 +170,9 @@ func (a *Azure) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
 	// UserDefinedRouting, which doesn't impose such constraints on secondary IPs.
 	loadBalancerBackendAddressPoolsArgument := (*networkInterface.IPConfigurations)[0].LoadBalancerBackendAddressPools
 	var attachedOutboundRule *network.SubResource
-	if (*networkInterface.IPConfigurations)[0].LoadBalancerBackendAddressPools != nil {
-	OuterLoop:
-		for _, ipconfig := range *networkInterface.IPConfigurations {
+OuterLoop:
+	for i, ipconfig := range *networkInterface.IPConfigurations {
+		if (*networkInterface.IPConfigurations)[i].LoadBalancerBackendAddressPools != nil {
```
ricky-rav, line 175: is there a reason why you're not using `ipconfig` here?

arghosh93: Sorry, my bad. I should have used ipconfig instead. I will correct this and push soon.
Thanks @ricky-rav for pointing this out.

ricky-rav:
/lgtm 
Thanks, Arnab!

ricky-rav:
@kyrtapz this should be good to go

PR #136, opened 2024-03-28
OCPBUGS-29915: GCP: Explicitly set UniverseDomain to the default universe domain

kyrtapz:
This is a follow up to https://github.com/openshift/cloud-network-config-controller/pull/132. 
Apparently it is not possible to set the universe domain with `option.WithUniverseDomain()`, this doesn't work because the base client used to determine whether there is a need to retrieve the  universe domain doesn't respect the value set through that option. 
To avoid this issue add the universe domain directly in the secret.

Tested this on my GCP cluster by removing the `universe_domain` field from the default secret.

### DIFF
--- a/pkg/cloudprovider/gcp.go
+++ b/pkg/cloudprovider/gcp.go
@@ -1,6 +1,7 @@
 package cloudprovider
 
 import (
+	"encoding/json"
 	"fmt"
 	"net"
 	"net/url"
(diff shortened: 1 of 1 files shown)


kyrtapz:
@martinkennelly ptal

martinkennelly:
/test e2e-azure-ovn

cluster init failure

PR #140, opened 2024-05-01
OCPBUGS-33154: Incorrect use of go wait groups

squirrd:
Minor change to the wait group to ensure the controller does not hang.  

The wg.Add(1) (line 198) should be moved closer to the function it is waiting for (line 226).

Having a function (that could exit) between wg.add() and the function with the matching wg.done() could lead to a state where the controller hangs, waiting for the wg.done() that is never called.


martinkennelly:
/test e2e-azure-ovn

Unrelated.

martinkennelly (approved):

kyrtapz:
/lgtm
Thanks @squirrd!

PR #144, opened 2024-05-15
OCPBUGS-33617: Avoid panic when looking up attachedOutboundRule.ID in azure

ricky-rav:
On a deployed cluster a panic happened here: https://github.com/openshift/cloud-network-config-controller/blob/0b191405af9a0632a09d9f857bb550c7efbba20f/pkg/cloudprovider/azure.go#L176

 It looks like the ID in `realPool.BackendAddressPoolPropertiesFormat.OutboundRule` might be `nil` in the pool we've just fetched: `realPool, err := a.getBackendAddressPool(*pool.ID)`. This is not at all expected, but let's check against nil pointers before dereferencing the ID.

### DIFF
--- a/pkg/cloudprovider/azure.go
+++ b/pkg/cloudprovider/azure.go
@@ -191,7 +191,7 @@ OuterLoop:
 						attachedOutboundRule = realPool.BackendAddressPoolPropertiesFormat.OutboundRule
 						break OuterLoop
 					}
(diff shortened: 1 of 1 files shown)


ricky-rav:
/cherry-pick release-4.15 release-4.14 release-4.13 release-4.12

PR #149, opened 2024-06-11
SDN-4773: Add support for Azure Managed Service Identity authentication

kyrtapz:
Add support for Azure Managed Service Identity authentication.
The PR also reorganizes the requirements around the azure credentials fields.

TODO:
The federated token file is no longer defaulted to the hardcoded path allowing the fallback to MSI.
Need to figure out whether that default value was ever used, if so we need to add a different mechanism to distinguish between workload identity and managed identity.
One option would be to set the `AZURE_FEDERATED_TOKEN_FILE` env in the CNO for self-hosted clusters only. It would mean that self-hosted clusters default to workload identity and hypershift clusters default to managed identity.

### DIFF
--- a/pkg/cloudprovider/azure.go
+++ b/pkg/cloudprovider/azure.go
@@ -51,50 +51,69 @@ type Azure struct {
 	azureWorkloadIdentityEnabled bool
 }
 
-func (a *Azure) initCredentials() error {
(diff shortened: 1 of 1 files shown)


kyrtapz:
/close
done in https://github.com/openshift/cloud-network-config-controller/pull/151

PR #151, opened 2024-08-05
SDN-4773: Add support for Azure Managed Service Identity authentication

bryan-cox:
This PR: 
* Adds support for Azure Managed Service Identity authentication
* Enables Azure MSI override for ARO HCP

### DIFF
--- a/pkg/cloudprovider/azure.go
+++ b/pkg/cloudprovider/azure.go
@@ -88,6 +88,10 @@ func (a *Azure) readAzureCredentialsConfig() (*azureCredentialsConfig, error) {
 	if err != nil {
 		klog.Infof("azure_federated_token_file not found in the secret: %v, falling back to AZURE_FEDERATED_TOKEN_FILE env", err)
 		cfg.tokenFile = os.Getenv("AZURE_FEDERATED_TOKEN_FILE")
+
+		if strings.TrimSpace(cfg.tokenFile) == "" {
+			cfg.tokenFile = "/var/run/secrets/openshift/serviceaccount/token"
+		}
 	}
 
 	cfg.subscriptionID, err = a.readSecretData("azure_subscription_id")
@@ -111,10 +115,6 @@ func (a *Azure) initCredentials() error {
 	cfg, err := a.readAzureCredentialsConfig()
 	if err != nil {
 		return err
-	}
-
-	if strings.TrimSpace(cfg.tokenFile) == "" {
-		cfg.tokenFile = "/var/run/secrets/openshift/serviceaccount/token"
 	}
 
 	a.resourceGroup = cfg.resourceGroup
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/azure.go
```diff
@@ -51,51 +51,70 @@ type Azure struct {
...
+
+	return &cfg, nil
+}
+func (a *Azure) initCredentials() error {
+	cfg, err := a.readAzureCredentialsConfig()
 	if err != nil {
-		tokenFile = os.Getenv("AZURE_FEDERATED_TOKEN_FILE")
-		if strings.TrimSpace(tokenFile) == "" {
-			// Use default value if no configuration is set
-			tokenFile = "/var/run/secrets/openshift/serviceaccount/token"
```
kyrtapz, line 95: With the new env var approach this defaulting should not be removed.

pkg/cloudprovider/azure.go
```diff
@@ -562,26 +581,65 @@ func (a *Azure) getAuthorizer(env azureapi.Environment, clientID, clientSecret,
...
+				},
+			}
+			if cfg.clientID != "" {
+				options.ID = azidentity.ClientID(cfg.clientID)
+			}
+			cred, err = azidentity.NewManagedIdentityCredential(&options)
+			if err != nil {
+				return nil, err
+			}
+		}
```
kyrtapz, line 631: This is no longer necessary with the new env var check above.

pkg/cloudprovider/azure.go
```diff
@@ -562,26 +581,65 @@ func (a *Azure) getAuthorizer(env azureapi.Environment, clientID, clientSecret,
...
 				Cloud: cloudConfig,
 			},
-			ClientID:      clientID,
-			TenantID:      tenantID,
-			TokenFilePath: tokenFile,
 		}
-		cred, err = azidentity.NewWorkloadIdentityCredential(&options)
+
+		var err error
+		cred, err = azidentity.NewManagedIdentityCredential(&options)
```
kyrtapz, line 595: Can you explain why isn't the clientID set even when it is provided?

kyrtapz:
There will need to be a CNO PR setting the `AZURE_MSI_AUTHENTICATION` env on the CNCC deployment.

pkg/cloudprovider/azure.go
bryan-cox: I defaulted it below. Please let me know if that's where you want that. I could also move it to `readAzureCredentialsConfig`.

pkg/cloudprovider/azure.go
bryan-cox: Yeah, Microsoft is building us an additional sidecar container that will take the MSI authentication call before it goes to IDMS, modify it a bit, approve it, and send back a token. Basically it will add info for the HC ID, mgmt cluster info, and the MSI client ID. 

So the operator itself doesn't need to know anything, just making the IDMS call and the sidecar container handles the rest of it.

bryan-cox:
> There will need to be a CNO PR setting the `AZURE_MSI_AUTHENTICATION` env on the CNCC deployment.

@kyrtapz It would just be this right? - https://github.com/openshift/cluster-network-operator/pull/2465

pkg/cloudprovider/azure.go
kyrtapz: please move it under https://github.com/openshift/cloud-network-config-controller/pull/151/files#diff-d3dbddd1556958367344cf05b435a98ffdc18ec9c060843fa1a5f4582d22f3b7R90. This is more clear when looking at how this var is set.

pkg/cloudprovider/azure.go
kyrtapz: I think we should remove it, this code is not reachable with AZURE_FEDERATED_TOKEN_FILE being defaulted to a constant path.

kyrtapz:
e2e-gcp-ovn is failing on image build.
This change is specific to azure and does not affect gcp.
/override e2e-gcp-ovn

PR #154, opened 2024-09-12
OCPBUGS-42004: Set the Managed Identity client ID

bryan-cox:
Set the Managed Identity client ID

### DIFF
--- a/pkg/cloudprovider/azure.go
+++ b/pkg/cloudprovider/azure.go
@@ -586,13 +586,14 @@ func (a *Azure) getAuthorizer(env azureapi.Environment, cfg *azureCredentialsCon
 		err  error
 	)
 
-	// MSI Override for ARO HCP
-	msi := os.Getenv("AZURE_MSI_AUTHENTICATION")
-	if msi == "true" {
+	// Managed Identity Override for ARO HCP
+	managedIdentityClientID := os.Getenv("ARO_HCP_MI_CLIENT_ID")
+	if managedIdentityClientID != "" {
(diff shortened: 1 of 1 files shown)


pkg/cloudprovider/azure.go
```diff
@@ -586,13 +586,14 @@ func (a *Azure) getAuthorizer(env azureapi.Environment, cfg *azureCredentialsCon
...
-	msi := os.Getenv("AZURE_MSI_AUTHENTICATION")
-	if msi == "true" {
+	// Managed Identity Override for ARO HCP
+	managedIdentityClientID := os.Getenv("ARO_HCP_MI_CLIENT_ID")
+	if managedIdentityClientID != "" {
 		options := azidentity.ManagedIdentityCredentialOptions{
 			ClientOptions: azcore.ClientOptions{
 				Cloud: cloudConfig,
 			},
+			ID: azidentity.ClientID(managedIdentityClientID),
```
fxierh, line 596: The mock-http demo seems to work without specifying the ID. Can this field be omitted?

bryan-cox: Someone I talked to from Microsoft said we would need a valid Client ID here when we are using the adapter sidecar containers. More info on that here - https://github.com/openshift/enhancements/pull/1659#issuecomment-2350008814.

bryan-cox:
/hold

Microsoft is requesting to not use the sidecar containers.

bryan-cox:
/close

An official decision was made on Monday to not use sidecars to authenticate with Azure. Closing out this PR.

PR #156, opened 2024-10-07
HOSTEDCP-2035: Use Client Cert Auth for ARO HCP deployments

bryan-cox:
Use Client Certificate Authentication for ARO HCP deployments. HyperShift will pass the needed environment variables for this authentication method: ARO_HCP_MI_CLIENT_ID, ARO_HCP_TENANT_ID, and ARO_HCP_CLIENT_CERTIFICATE_PATH.

### DIFF
--- a/pkg/cloudprovider/azure.go
+++ b/pkg/cloudprovider/azure.go
@@ -20,6 +20,7 @@ import (
 	"github.com/jongio/azidext/go/azidext"
 	v1 "github.com/openshift/api/cloudnetwork/v1"
 	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/cloud-network-config-controller/pkg/filewatcher"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
@@ -586,17 +587,37 @@ func (a *Azure) getAuthorizer(env azureapi.Environment, cfg *azureCredentialsCon
 		err  error
 	)
 
-	// MSI Override for ARO HCP
-	msi := os.Getenv("AZURE_MSI_AUTHENTICATION")
-	if msi == "true" {
-		options := azidentity.ManagedIdentityCredentialOptions{
+	// Managed Identity Override for ARO HCP
+	managedIdentityClientID := os.Getenv("ARO_HCP_MI_CLIENT_ID")
+	if managedIdentityClientID != "" {
+		klog.Info("Using client certification Azure authentication for ARO HCP")
+		options := &azidentity.ClientCertificateCredentialOptions{
 			ClientOptions: azcore.ClientOptions{
 				Cloud: cloudConfig,
 			},
+			SendCertificateChain: true,
 		}
 
-		var err error
-		cred, err = azidentity.NewManagedIdentityCredential(&options)
+		tenantID := os.Getenv("ARO_HCP_TENANT_ID")
+		certPath := os.Getenv("ARO_HCP_CLIENT_CERTIFICATE_PATH")
+
+		certData, err := os.ReadFile(certPath)
+		if err != nil {
+			return nil, fmt.Errorf(`failed to read certificate file "%s": %v`, certPath, err)
+		}
+
+		certs, key, err := azidentity.ParseCertificates(certData, []byte{})
+		if err != nil {
+			return nil, fmt.Errorf(`failed to parse certificate data "%s": %v`, certPath, err)
+		}
+
+		// Watch the certificate for changes; if the certificate changes, the pod will be restarted
+		err = filewatcher.WatchFileForChanges(certPath)
+		if err != nil {
+			return nil, err
+		}
+
+		cred, err = azidentity.NewClientCertificateCredential(tenantID, managedIdentityClientID, certs, key, options)
 		if err != nil {
 			return nil, err
 		}
--- a/pkg/filewatcher/filewatcher.go
+++ b/pkg/filewatcher/filewatcher.go
@@ -1,0 +1,68 @@
+package filewatcher
+
+import (
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/fsnotify/fsnotify"
+	"k8s.io/klog/v2"
+)
+
+var watchCertificateFileOnce sync.Once
+
+// WatchFileForChanges watches the file, fileToWatch, for changes. If the file contents have changed, the pod this
+// function is running on will be restarted.
+func WatchFileForChanges(fileToWatch string) error {
+	var err error
+
+	// This starts only one occurrence of the file watcher, which watches the file, fileToWatch.
+	watchCertificateFileOnce.Do(func() {
+		klog.Infof("Starting the file change watcher on file, %s", fileToWatch)
+
+		// Update the file path to watch in case this is a symlink
+		fileToWatch, err = filepath.EvalSymlinks(fileToWatch)
+		if err != nil {
+			return
+		}
+		klog.Infof("Watching file, %s", fileToWatch)
+
+		// Start the file watcher to monitor file changes
+		go func() {
+			err := checkForFileChanges(fileToWatch)
+			klog.Errorf("Error checking for file changes: %v", err)
+		}()
+	})
+	return err
+}
+
+// checkForFileChanges starts a new file watcher. If the file is changed, the pod running this function will exit.
+func checkForFileChanges(path string) error {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return err
+	}
+
+	go func() {
+		for {
+			select {
+			case event, ok := <-watcher.Events:
+				if ok && (event.Has(fsnotify.Write) || event.Has(fsnotify.Chmod) || event.Has(fsnotify.Remove)) {
+					klog.Infof("file, %s, was modified, exiting...", event.Name)
+					os.Exit(0)
+				}
+			case err, ok := <-watcher.Errors:
+				if ok {
+					klog.Errorf("file watcher error: %v", err)
+				}
+			}
+		}
+	}()
+
+	err = watcher.Add(path)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.3.0
 	github.com/Azure/go-autorest/autorest v0.11.27
 	github.com/aws/aws-sdk-go v1.37.8
+	github.com/fsnotify/fsnotify v1.7.0
 	github.com/google/uuid v1.6.0
 	github.com/gophercloud/gophercloud v0.25.1-0.20220718160629-0721d75e876f
 	github.com/gophercloud/utils v0.0.0-20220307143606-8e7800759d16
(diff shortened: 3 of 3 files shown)


pkg/cloudprovider/azure.go
```diff
@@ -586,17 +589,47 @@ func (a *Azure) getAuthorizer(env azureapi.Environment, cfg *azureCredentialsCon
...
+		if err != nil {
+			return nil, fmt.Errorf(`failed to read certificate file "%s": %v`, certPath, err)
+		}
+
+		certs, key, err := azidentity.ParseCertificates(certData, []byte{})
+		if err != nil {
+			return nil, fmt.Errorf(`failed to parse certificate data "%s": %v`, certPath, err)
+		}
+
+		// Set up a watch on our config file; if it changes, we should exit -
```
kyrtapz, line 616: config file? it is meant to exit if the cert changes right?
More broadly, is there a plan to add cert reload capability to the ARO go client?

pkg/cloudprovider/azure.go
```diff
@@ -586,17 +589,47 @@ func (a *Azure) getAuthorizer(env azureapi.Environment, cfg *azureCredentialsCon
...
+		// (we don't have the ability to dynamically reload config changes).
+		stopCh := make(chan struct{})
+		err = interrupt.New(func(s os.Signal) {
+			_, _ = fmt.Fprintf(os.Stderr, "interrupt: Gracefully shutting down ...\n")
```
kyrtapz, line 620: What is the purpose of logging like that?

pkg/cloudprovider/azure.go
```diff
@@ -586,17 +589,47 @@ func (a *Azure) getAuthorizer(env azureapi.Environment, cfg *azureCredentialsCon
...
+			close(stopCh)
+		}).Run(func() error {
+			if err := watchForChanges(certPath, stopCh); err != nil {
```
kyrtapz, line 623:
```suggestion
			return watchForChanges(certPath, stopCh)
```

pkg/cloudprovider/azure.go
```diff
@@ -586,17 +589,47 @@ func (a *Azure) getAuthorizer(env azureapi.Environment, cfg *azureCredentialsCon
...
-		options := azidentity.ManagedIdentityCredentialOptions{
+	// Managed Identity Override for ARO HCP
+	managedIdentityClientID := os.Getenv("ARO_HCP_MI_CLIENT_ID")
+	if managedIdentityClientID != "" {
+		klog.Info("Using client certification Azure authentication for ARO HCP")
+		options := &azidentity.ClientCertificateCredentialOptions{
 			ClientOptions: azcore.ClientOptions{
 				Cloud: cloudConfig,
 			},
+			SendCertificateChain: true,
```
kyrtapz, line 600: Out of curiosity, could you comment on why is this required?

pkg/cloudprovider/azure.go
```diff
@@ -658,3 +691,59 @@ func (a *Azure) getNodeLock(nodeName string) *sync.Mutex {
...
 	return id[strings.LastIndex(id, "/"):]
 }
+
+// watchForChanges closes stopCh if the configuration file changed.
+func watchForChanges(configPath string, stopCh chan struct{}) error {
+	configPath, err := filepath.Abs(configPath)
+	if err != nil {
+		return err
+	}
+	watcher, err := fsnotify.NewWatcher()
```
kyrtapz, line 701: Missing `defer watcher.Close()`

pkg/cloudprovider/azure.go
```diff
@@ -658,3 +691,59 @@ func (a *Azure) getNodeLock(nodeName string) *sync.Mutex {
 func getNameFromResourceID(id string) string {
 	return id[strings.LastIndex(id, "/"):]
 }
+
+// watchForChanges closes stopCh if the configuration file changed.
```
kyrtapz, line 695: NIT: this function is not specific to config files, can you update the comment to better reflect what it does?

pkg/cloudprovider/azure.go
```diff
@@ -658,3 +691,59 @@ func (a *Azure) getNodeLock(nodeName string) *sync.Mutex {
...
+				return err
+			}
+		} else {
+			break
+		}
+	}
+	go func() {
+		for {
+			select {
+			case <-stopCh:
```
kyrtapz, line 730: shouldn't this cause a return?

pkg/cloudprovider/azure.go
```diff
@@ -658,3 +691,59 @@ func (a *Azure) getNodeLock(nodeName string) *sync.Mutex {
...
+				}
+				klog.V(2).Infof("Configuration file %s changed, exiting...", event.Name)
+				close(stopCh)
+				return
+			case err, ok := <-watcher.Errors:
+				if !ok {
+					klog.Errorf("failed to watch config file %s", p)
+					return
+				}
+				klog.Errorf("fsnotify error: %v", err)
```
kyrtapz, line 744: What will happen in this case? We will just stop watching for cert changes? shouldn't this close the channel?

pkg/cloudprovider/azure.go
```diff
@@ -586,17 +589,47 @@ func (a *Azure) getAuthorizer(env azureapi.Environment, cfg *azureCredentialsCon
...
+		if err != nil {
+			return nil, fmt.Errorf(`failed to parse certificate data "%s": %v`, certPath, err)
+		}
+
+		// Set up a watch on our config file; if it changes, we should exit -
+		// (we don't have the ability to dynamically reload config changes).
+		stopCh := make(chan struct{})
+		err = interrupt.New(func(s os.Signal) {
+			_, _ = fmt.Fprintf(os.Stderr, "interrupt: Gracefully shutting down ...\n")
+			close(stopCh)
```
kyrtapz, line 621: I am confused about what actually happens when stopCh is closed and why is it closed here?

pkg/cloudprovider/azure.go
bryan-cox: Yeah this is meant to exit (i.e. reboot the pod) when the cert changes. Azure is planning to add a reload capability to its SDK but it is not there yet.

pkg/cloudprovider/azure.go
bryan-cox: I think this was just a copy/paste thing from somewhere. Happy to change this to klog.

pkg/cloudprovider/azure.go
bryan-cox: This is related to 1st party Azure technology but with this, the certificate authentication doesn't require pinning certificates to applications, basically saying "Trust this certificate subject issued from this CA". cc @bennerv

pkg/cloudprovider/azure.go
bryan-cox: Yeah I'll make it more generic

pkg/cloudprovider/azure.go
bryan-cox: Yeah I believe so, we would stop watching for file changes. The channel should be closed on L621.

pkg/cloudprovider/azure.go
bryan-cox: As I understand it, the caller of `watchForChanges` is ultimately responsible for determining if the channel needs to be closed or what needs done with it. So when `watchForChanges` returns it would pass the return back up the stack to `Run()` which passes it back to `New()`.

I copied this code from an example @enxebre put together, but it's essentially the same method used by OpenShift SDN [here](https://github.com/openshift/sdn/blob/18a57d7da14c7a45c4e8d55254bf7a89cee0ac43/pkg/cmd/openshift-sdn-node/cmd.go#L89).

pkg/cloudprovider/azure.go
bryan-cox: No I don't believe so. I think we only want to return if there were changes or an error.

pkg/cloudprovider/azure.go
bennerv: Note that it has no averse affect if you don't use SNI authentication with the 1p Azure technology.  The azure-sdk-for-go comment has a [bit more detail:](https://github.com/Azure/azure-sdk-for-go/blob/9591fab0ac0e5e3e2707d6db579331b136fb00ea/sdk/azidentity/client_certificate_credential.go#L45-L48) 
> 	// SendCertificateChain controls whether the credential sends the public certificate chain in the x5c
>	// header of each token request's JWT. This is required for Subject Name/Issuer (SNI) authentication.
>	// Defaults to False.

pkg/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,75 @@
...
+	go func() {
+		for {
+			select {
+			case <-done:
+				return
+			case <-ticker.C:
+				klog.Infof("Checking file for changes, %s", fileToWatch)
+				fileContents, err := os.ReadFile(fileToWatch)
+				if err != nil {
+					klog.Error("failed to read the file: %v", err)
```
kyrtapz, line 51: When a file read fails we will just stop watching it and if it changes this will never be caught.

pkg/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,75 @@
...
+		go checkForFileChanges(fileToWatch, interval)
+	})
+	return err
+}
+
+func checkForFileChanges(fileToWatch string, interval time.Duration) {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	done := make(chan bool)
```
kyrtapz, line 40: What's the point of this channel?

kyrtapz (changes requested):
I am not confident in the new approach, It causes a hard stop and has a potentially significant delay.

pkg/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,75 @@
...
+
+	watchCertificateFileOnce.Do(func() {
+		klog.Infof("Starting the file change watcher on file, %s", fileToWatch)
+		fileContents, err = os.ReadFile(fileToWatch)
+		if err != nil {
+			err = fmt.Errorf("failed to read file on initialization: %v", err)
+			return
+		}
+
+		initialFileHash = hashSimple(fileContents)
```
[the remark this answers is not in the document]
bryan-cox: We can certainly poll sooner than 30m for the file check. How quick would you like it to poll?

pkg/filewatcher/filewatcher.go
```diff
@@ -0,0 +1,75 @@
...
+	return err
+}
+
+func checkForFileChanges(fileToWatch string, interval time.Duration) {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	done := make(chan bool)
+
+	go func() {
```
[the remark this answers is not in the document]
bryan-cox: Yeah, I didn't need the second goroutine so I removed it.

pkg/filewatcher/filewatcher.go
bryan-cox: The return should just return from the sync.Once function and the error would be returned at the end of `WatchFileForChanges`

pkg/filewatcher/filewatcher.go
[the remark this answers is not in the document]
bryan-cox: The reason for the periodic hash check - the previous file watcher code was not catching the changes when I was testing this function against the CPO. I could exec into the pod and see the file changed but the pod was not restarted nor did I see any messages about the file changing.

This current change seemed just as simple and required less code. This change was working when I was testing the same function against the CPO.

pkg/filewatcher/filewatcher.go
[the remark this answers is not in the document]
bryan-cox: Apparently this is needed. The pod does not restart when I removed this.

PR #159, opened 2024-11-25
[release-4.17] OCPBUGS-44861: Use Client Cert Auth for ARO HCP deployments

bryan-cox:
Use Client Certificate Authentication for ARO HCP deployments. HyperShift will pass the needed environment variables for this authentication method: ARO_HCP_MI_CLIENT_ID, ARO_HCP_TENANT_ID, and ARO_HCP_CLIENT_CERTIFICATE_PATH.


bryan-cox:
/close

Closing for now since it looks like ARO HCP will private preview with 4.18.0 rather than 4.17.z.

PR #162, opened 2025-02-20
OCPBUGS-48111: Fix pool name dereference for azure

ricky-rav:


pkg/cloudprovider/azure.go
```diff
@@ -207,7 +207,11 @@ OuterLoop:
 				// attached to it
 				realPool, err := a.getBackendAddressPool(*pool.ID)
 				if err != nil {
-					return fmt.Errorf("error looking up backend address pool %s with ID %s: %v", *pool.Name, *pool.ID, err)
+					poolName := ""
+					if pool.Name != nil {
+						poolName = *pool.Name
+					}
+					return fmt.Errorf("error looking up backend address pool %s with ID %s: %v", poolName, *pool.ID, err)
```
tssurya, line 214: hmmm so if the pool.Name is empty then this error is going to print an empty name ? :D that's not going to be useful is it? might as well not include that info in the error if its nil.. 
I am not sure I understand when the pool.Name can be nil when pool.ID is not nil..

ricky-rav: ah it will print the name only if there is one and it will always print the pool ID.
From the customer case, the customer ended up with an empty pool name when the backend address pool didn't have read permission :) 

> CredentialsRequest openshift-cloud-network-config-controller-azure is missing permission: Microsoft.Network/load-balancers/backendAddressPools/read
> 
> Customer Has Confirmation that manually appending this permission to the workload identity has resolved the panic..

Timdawson264: https://github.com/openshift/cluster-network-operator/blob/master/manifests/02-cncc-credentials.yaml#L66 please dont forget to add the missing permission.  I'm not sure how getBackendAddressPool ever worked without it.

ricky-rav:
Closing this PR, as the issue was fixed by:

- https://github.com/openshift/cloud-network-config-controller/commit/16c498c090ce122c87f346eec7201b26b064868b
- https://github.com/openshift/cluster-network-operator/pull/2684

PR #163, opened 2025-02-26
SDN-5690:Rebase k8s to 1.32

asood-rh:
go get -u github.com/openshift/api
go get -u github.com/openshift/client-go
go get -u github.com/openshift/library-go
go get -u k8s.io/api
go get -u k8s.io/apimachinery
go get -u k8s.io/client-go
go get -u k8s.io/cloud-provider
go get -u k8s.io/klog/v2
go get -u k8s.io/utils
go get -u sigs.k8s.io/controller-runtime
go mod tidy; go mod vendor

/cc @jluhrsen 

Fix the build error in main.go to pass in passiveClock argument of type "k8s.io/utils/clock".PassiveClock


jluhrsen:
@asood-rh , just FYI that I have done a similar thing in the past and I was asked to squash the "fix" changes in to the same commit as the rebase update changes. the reason was that we didn't want to have a commit that wouldn't build which is
what we currently have with just the rebase changes.

of course, you should add the error and fix to the commit message (and GH description) to help reviewers understand
what's going one more easily.

asood-rh:
@jluhrsen Who could review this PR? All tests except security test passed.

jluhrsen:
> @jluhrsen Who could review this PR? All tests except security test passed.

@asood-rh , we need someone from the approvers section of [this OWNERS file](https://github.com/openshift/cloud-network-config-controller/blob/master/OWNERS). cc @trozet @abhat @kyrtapz @danwinship

cmd/cloud-network-config-controller/main.go
```diff
@@ -297,7 +299,7 @@ func createFeatureGateAccessor(ctx context.Context, cfg *rest.Config, operatorNa
 		Kind:       "Deployment",
 		Namespace:  namespace,
 		Name:       deploymentName,
-	})
+	}, passiveClock)
```
kyrtapz, line 302: I don't think this will work, passiveClock is an empty interface so when CNCC will try to generate an event the code will likely panic.
To solve this you could just use `clock.RealClock`:
```suggestion
	}, clock.RealClock{})
```

kyrtapz (changes requested):

PR #166, opened 2025-03-26
CNTRLPLANE-112: Add new Azure authentication type for managed Azure HCP for cloud-network-config-controller

bryan-cox:
This PR: 
- adds the new Azure authentication for managed Azure HCP called UserAssignedIdentityCredentials.
- refactors several items to use new Azure SDK for go libraries

See [Persisting MSI Credentials](https://docs.google.com/document/d/1tX2avo8ORyjqL4GJNwxq-wQTM3-LC7zBTP5V9EgwU74/edit?tab=t.0#heading=h.12bfa2iapy8v) for more info.


bryan-cox:
Needs https://github.com/openshift/hypershift/pull/5903 for there to be no errors in the CNCC log for Azure HCP.

pkg/cloudprovider/azure.go
```diff
@@ -370,9 +371,12 @@ func (a *Azure) getSubnet(networkInterface armnetwork.Interface) (*net.IPNet, *n
 	}
 	var v4Subnet, v6Subnet *net.IPNet
 	for _, addressPrefix := range addressPrefixes {
-		_, subnet, err := net.ParseCIDR(addressPrefix)
+		if addressPrefix == nil {
+			return nil, nil, fmt.Errorf("error retrieving associated address prefix, err: %v", err)
```
kyrtapz, line 372: What is the `err` here?

pkg/cloudprovider/azure.go
```diff
@@ -550,30 +554,30 @@ func (a *Azure) getAddressPrefixes(networkInterface armnetwork.Interface) ([]str
...
 		return nil, fmt.Errorf("error retrieving subnet IP configuration, err: %v", err)
 	}
 	// Check the list of subnets first. If a subnet with the subnet name is found, then use that
 	// instead of virtualNetwork.AddressSpace.AddressPrefixes which only contains the main subnet's
 	// address prefix.
 	// FIXME: This might not work for IPv6.
-	if virtualNetwork.Subnets != nil {
-		for _, vns := range *virtualNetwork.Subnets {
-			if vns.Name != nil && vns.AddressPrefix != nil &&
+	if virtualNetwork.Properties.Subnets != nil {
```
kyrtapz, line 562: `virtualNetwork.Properties` is a pointer, please check for nil first.

pkg/cloudprovider/azure.go
```diff
@@ -134,27 +136,36 @@ func (a *Azure) initCredentials() error {
...
-	authorizer, err := a.getAuthorizer(a.env, cfg)
+	cred, cloudConfig, err := a.getAzureCredentials(a.env, cfg)
 	if err != nil {
 		return err
 	}
 
-	a.vmClient = compute.NewVirtualMachinesClientWithBaseURI(a.env.ResourceManagerEndpoint, cfg.subscriptionID)
-	a.vmClient.Authorizer = authorizer
-	_ = a.vmClient.AddToUserAgent(UserAgent)
+	options := &arm.ClientOptions{
```
kyrtapz, line 144: I am confused by this, all creds returned from `getAzureCredentials` already have the clientOptions set to this value. What's the point of doing it again here? Or is the one in getAzureCredentials redundant?

kyrtapz (changes requested):
Thanks @bryan-cox 
Please add nil de-reference checks where applicable. 
The azure-serial job look good but the aks one doesn't: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cloud-network-config-controller/166/pull-ci-openshift-cloud-network-config-controller-master-hypershift-e2e-aks/1904907670675001344/artifacts/hypershift-e2e-aks/hypershift-azure-run-e2e/artifacts/TestCreateCluster/namespaces/e2e-clusters-6nsr9-create-cluster-92k5f/core/pods/logs/cloud-network-config-controller-5df5674d45-d555q-controller.log is that expected because of the missing HyperShift changes?

pkg/cloudprovider/azure.go
bryan-cox: Ah, that's a typo. That will be removed

pkg/cloudprovider/azure.go
bryan-cox: Part of this changed in a newer commit from when this comment was made. Right now, I'm reusing the same cloudConfig generated in `getAzureCredentials`.  All of the clients have an options parameter so I'm making sure we are using the same cloudConfig in those as we used in `getAzureCredentials`.

kyrtapz: It is still unclear why are we using the same client options in creds and as a parameter here but I think it is fine for now.

pkg/cloudprovider/azure.go
```diff
@@ -173,14 +184,14 @@ func (a *Azure) AssignPrivateIP(ip net.IP, node *corev1.Node) error {
 	if err != nil {
 		return err
 	}
-	applicationSecurityGroups := (*networkInterfaces[0].IPConfigurations)[0].InterfaceIPConfigurationPropertiesFormat.ApplicationSecurityGroups
+	applicationSecurityGroups := networkInterfaces[0].Properties.IPConfigurations[0].Properties.ApplicationSecurityGroups
```
kyrtapz, line 187: Can you add a nil check for properties?

pkg/cloudprovider/azure.go
```diff
@@ -228,25 +239,25 @@ OuterLoop:
...
+			PrivateIPAllocationMethod:       ptr.To(armnetwork.IPAllocationMethodStatic),
+			Subnet:                          networkInterface.Properties.IPConfigurations[0].Properties.Subnet,
 			Primary:                         &untrue,
 			LoadBalancerBackendAddressPools: loadBalancerBackendAddressPoolsArgument,
 			ApplicationSecurityGroups:       applicationSecurityGroups,
 		},
 	}
 	for _, ipCfg := range ipConfigurations {
-		if ipCfg.PrivateIPAddress != nil && *ipCfg.PrivateIPAddress == ipc {
+		if ipCfg.Properties.PrivateIPAddress != nil && ptr.Deref(ipCfg.Properties.PrivateIPAddress, "") == ipc {
```
kyrtapz, line 260:
```suggestion
		if ptr.Deref(ipCfg.Properties.PrivateIPAddress, "") == ipc {
```

pkg/cloudprovider/azure.go
```diff
@@ -256,13 +267,13 @@ OuterLoop:
 		}
 	}
 	ipConfigurations = append(ipConfigurations, newIPConfiguration)
-	networkInterface.IPConfigurations = &ipConfigurations
+	networkInterface.Properties.IPConfigurations = ipConfigurations
 	// Send the request
-	result, err := a.createOrUpdate(networkInterface)
+	err = a.createOrUpdate(networkInterface)
```
kyrtapz, line 272:
```suggestion
	return a.createOrUpdate(networkInterface)
```

pkg/cloudprovider/azure.go
```diff
@@ -283,10 +294,10 @@ func (a *Azure) ReleasePrivateIP(ip net.IP, node *corev1.Node) error {
 	// following the order Azure specifies.
 	networkInterface := networkInterfaces[0]
 	// Release the IP
-	keepIPConfiguration := []network.InterfaceIPConfiguration{}
+	keepIPConfiguration := []*armnetwork.InterfaceIPConfiguration{}
 	ipAssigned := false
-	for _, ipConfiguration := range *networkInterface.IPConfigurations {
-		if assignedIP := net.ParseIP(*ipConfiguration.PrivateIPAddress); assignedIP != nil && !assignedIP.Equal(ip) {
+	for _, ipConfiguration := range networkInterface.Properties.IPConfigurations {
```
kyrtapz, line 299: Please add a check for nil in `networkInterface.Properties`

pkg/cloudprovider/azure.go
```diff
@@ -296,13 +307,13 @@ func (a *Azure) ReleasePrivateIP(ip net.IP, node *corev1.Node) error {
 	if !ipAssigned {
 		return NonExistingIPError
 	}
-	networkInterface.IPConfigurations = &keepIPConfiguration
+	networkInterface.Properties.IPConfigurations = keepIPConfiguration
 	// Send the request
-	result, err := a.createOrUpdate(networkInterface)
+	err = a.createOrUpdate(networkInterface)
```
kyrtapz, line 312:
```suggestion
	return a.createOrUpdate(networkInterface)
```

pkg/cloudprovider/azure.go
```diff
@@ -339,29 +350,30 @@ func (a *Azure) GetNodeEgressIPConfiguration(node *corev1.Node, cloudPrivateIPCo
 	return []*NodeEgressIPConfiguration{config}, nil
 }
 
-func (a *Azure) createOrUpdate(networkInterface network.Interface) (network.InterfacesCreateOrUpdateFuture, error) {
+func (a *Azure) createOrUpdate(networkInterface armnetwork.Interface) error {
 	ctx, cancel := context.WithTimeout(a.ctx, defaultAzureOperationTimeout)
 	defer cancel()
-	return a.networkClient.CreateOrUpdate(ctx, a.resourceGroup, *networkInterface.Name, networkInterface)
-}
+	_, err := a.networkClient.BeginCreateOrUpdate(ctx, a.resourceGroup, ptr.Deref(networkInterface.Name, ""), networkInterface, nil)
```
kyrtapz, line 356: so you removed the `a.waitForCompletion(result)` from all of the calls to `a.waitForCompletion(result)`.
This means to me that we are no longer waiting for the operation to complete, why?
`BeginCreateOrUpdate` returns a poller that could help achieve the same behavior as we had and I think we should use it here.

pkg/cloudprovider/azure.go
```diff
@@ -339,29 +350,30 @@ func (a *Azure) GetNodeEgressIPConfiguration(node *corev1.Node, cloudPrivateIPCo
...
 	if err != nil {
 		return nil, nil, fmt.Errorf("error retrieving associated address prefix for network interface, err: %v", err)
 	}
 	var v4Subnet, v6Subnet *net.IPNet
 	for _, addressPrefix := range addressPrefixes {
-		_, subnet, err := net.ParseCIDR(addressPrefix)
+		if addressPrefix == nil {
+			return nil, nil, fmt.Errorf("error retrieving associated address prefix")
+		}
+		_, subnet, err := net.ParseCIDR(ptr.Deref(addressPrefix, ""))
```
kyrtapz, line 374: what is the point of `ptr.Deref` if you checked for nil above?

pkg/cloudprovider/azure.go
```diff
@@ -399,36 +411,36 @@ func (a *Azure) getCapacity(networkInterface network.Interface, cloudPrivateIPsC
...
-	return &instance, nil
+	return &instance.VirtualMachine, nil
 }
 
 // getNetworkInterfaces returns a slice of network.Interface with the
 // primary one first, if it exists, else in the order assigned by Azure.
-func (a *Azure) getNetworkInterfaces(instance *compute.VirtualMachine) ([]network.Interface, error) {
-	if instance.NetworkProfile == nil {
+func (a *Azure) getNetworkInterfaces(instance *armcompute.VirtualMachine) ([]armnetwork.Interface, error) {
+	if instance.Properties.NetworkProfile == nil {
```
kyrtapz, line 431: Please add a check for `instance.Properties` being nil.

pkg/cloudprovider/azure.go
```diff
@@ -437,9 +449,9 @@ func (a *Azure) getNetworkInterfaces(instance *compute.VirtualMachine) ([]networ
 		}
 	}
 	// Get the rest and append that.
-	for _, netif := range *instance.NetworkProfile.NetworkInterfaces {
-		if netif.NetworkInterfaceReferenceProperties != nil && ((netif.Primary != nil && !*netif.Primary) || netif.Primary == nil) {
-			intf, err := a.getNetworkInterface(*netif.ID)
+	for _, netif := range instance.Properties.NetworkProfile.NetworkInterfaces {
+		if netif.Properties != nil && ((netif.Properties.Primary != nil && !ptr.Deref(netif.Properties.Primary, false)) || netif.Properties.Primary == nil) {
```
kyrtapz, line 453: I can this can be simplified to
```suggestion
		if netif.Properties != nil && !ptr.Deref(netif.Properties.Primary, false) {
```

kyrtapz (commented):
Thanks, looking better but there are still a few places where you are missing nil deref checks on the `Properties` subfield.

kyrtapz:
@bryan-cox have you seen my second round comments? 
As for the azure-serial job I don't think the failure is related to this change, as long as the egressIP tests are passing should be good.

kyrtapz:
@huiran0826 this is the CNCC PR that we can start pre-merge testing, we need to ensure that egressIP works as expected for AKS. Please work with @bryan-cox to get a cluster where these changes are applied, it might require an additional hypershift PR.

pkg/cloudprovider/azure.go
bryan-cox: Didn't add this since [this comment](https://github.com/openshift/cloud-network-config-controller/pull/166/commits/8d7fa811c2b01fc6e6cb794d1593745c3fdd4536#r2028300777) changes what is requested in this comment.

pkg/cloudprovider/azure.go
bryan-cox: Didn't add this since [this comment](https://github.com/openshift/cloud-network-config-controller/pull/166/commits/8d7fa811c2b01fc6e6cb794d1593745c3fdd4536#r2028300777) changes what is requested in this comment.

pkg/cloudprovider/azure.go
bryan-cox: There is not, but I was just cleaning them all up from earlier PR comments. I will remove it since you pointed it out.

PR #167, opened 2025-03-27
OSASINFRA-3755: Prefer CA cert from credentials secret

stephenfin:
In openshift/cloud-credential-operator/pull/780, we have added the ability for `cloud-credential-operator` to consume a CA cert from the root credentials secret and to include in the credentials secrets it provisions.
In openshift/installer/pull/9194, we have modified the Installer to start setting this field where necessary.

Adapt `cloud-network-config-operator` to allow it to start consuming the CA cert from this place. We maintain fallbacks for the previous locations of the cert for now, but these can be removed in the next release.

### DIFF
--- a/pkg/cloudprovider/openstack.go
+++ b/pkg/cloudprovider/openstack.go
@@ -119,9 +119,24 @@ func (o *OpenStack) initCredentials() error {
 	}
 
 	// Read CA information - needed for self-signed certificates.
-	// That information is stored in ConfigMap kube-cloud-config.
-	caBundle := filepath.Join(o.cfg.ConfigDir, "ca-bundle.pem")
+	// That information is stored in ConfigMap kube-cloud-config and, since 4.19,
+	// in the secret cloud-credentials. Prefer the latter if found.
+	caBundle := filepath.Join(o.cfg.CredentialDir, "cacert")
 	userCACert, err := os.ReadFile(caBundle)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("could not parse file '%s', err: %q", caBundle, err)
+		}
+
+		// Fallback for retrieving CA cert from the CCM config. Starting in
+		// OCP 4.19, cloud-credential-operator provides this in the credential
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/openstack.go
```diff
@@ -119,9 +119,24 @@ func (o *OpenStack) initCredentials() error {
 	}
 
 	// Read CA information - needed for self-signed certificates.
-	// That information is stored in ConfigMap kube-cloud-config.
-	caBundle := filepath.Join(o.cfg.ConfigDir, "ca-bundle.pem")
-	userCACert, err := os.ReadFile(caBundle)
+	// That information is stored in ConfigMap kube-cloud-config and, since 4.19,
+	// in the secret cloud-credentials. Prefer the latter if found.
+	caBundle := filepath.Join(o.cfg.CredentialDir, "cacert")
+	userCACert, err := os.ReadFile(clientConfigFile)
```
mandre, line 125:
```suggestion
	userCACert, err := os.ReadFile(caBundle)
```

mandre (commented):
/lgtm

kyrtapz:
I this expected in CI?
```
I0401 17:50:18.242251       1 openstack.go:154] Could not find custom CA bundle in file '/kube-cloud-config/ca-bundle.pem' - some environments require a custom CA to work correctly
```
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_cloud-network-config-controller/167/pull-ci-openshift-cloud-network-config-controller-master-e2e-openstack-ovn-serial-e2e-only/1907114613669892096/artifacts/e2e-openstack-ovn-serial-e2e-only/gather-extra/artifacts/pods/openshift-cloud-network-config-controller_cloud-network-config-controller-7d67f9679d-sfjv2_controller.log

If I read the code correctly it means that either both CAs do not exist or the one that was read is empty. 
Additionally, maybe it would be good to add a log highlighting that the fallback is taking place?

PR #168, opened 2025-04-11
OCPBUGS-45891: Increase API call timeout to 30 second

arghosh93:
This PR is to increase client timeout for any API call to 30 second. A warning message gets logged if response from API is not received before previously set timeout period gets over.

### DIFF
--- a/pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
+++ b/pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
```diff
@@ -504,6 +516,9 @@ func (c *CloudPrivateIPConfigController) getCloudPrivateIPConfig(name string) (*
 		}
 		return nil, err
 	}
+	if time.Until(warningTime) <= 0 {
+		klog.Warningf("CloudPrivateIPConfig: Get API call took longer than expected for resource %q, Egress IP configuration might be delayed", name)
+	}
```
tkashem, line 521: logging the duration of the API call will be helpful

tkashem (commented):
LGTM (have a minor coment)

PR #173, opened 2025-05-21
OCPBUGS-45056: Azure: Don't add EgressIP to LB backends

mdbooth:
**DO NOT MERGE**

This solution assumes that the lack of external connectivity for EgressIPs is a known limitation and therefore deliberately makes no attempt to preserve it. However, it turns out this is a regression rather than a known limitation, and one that only affects clusters installed since the regression was introduced. Therefore by removing external connectivity in all cases, this PR would itself cause a regression for users who were not otherwise affected by it because they had upgraded from an unaffected version.

See https://issues.redhat.com/browse/OCPBUGS-57422 for discussion of the regression.

Leaving it open for now, but I (@mdbooth) expect to close this without merging.

---
With this change we no longer set LoadBalancerBackendAddressPools on IPConfigurations associated with an EgressIP. We do this for 2 reasons:
* We do not require the LoadBalancer to direct traffic to the EgressIP for inbound traffic.
* Adding these rules multiplies the number of loadbalancer rules on a NIC by the number of Egress IPs assigned to that NIC

The latter will prevent the addition of new EgressIPs and LoadBalancer services which can be added to the whole cluster once any NIC in the cluster exceeds 300 rules.

**NOTE**: Egress IPs were previously already excluded from any LoadBalancerBackendAddressPool with an outbound rule. This means that any pod matched by an Egress IP was *already prevented from making outbound connections*. This change was made in https://github.com/openshift/cloud-network-config-controller/pull/121. There is currently no solution to this issue, except possibly using a NAT gateway instead.

The snyk changes are unrelated, but required to get `test-security` to pass.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -7,6 +7,7 @@ exclude:
     # https://issues.redhat.com/browse/SDN-4462.
     # ** KEEP THIS LIST SORTED FOR CONSISTENCY **
     - vendor/github.com/aws/aws-sdk-go/aws/credentials/ssocreds/provider.go
+    - vendor/github.com/aws/aws-sdk-go/aws/endpoints/dep_service_ids.go
     - vendor/github.com/aws/aws-sdk-go/service/sts/api.go
     - vendor/github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/local/server.go
     - vendor/github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/accesstokens/accesstokens.go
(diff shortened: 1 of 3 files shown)


mdbooth:
/hold

We should consider the upgrade here. Ideally we will update existing NICs to remove this configuration immediately on upgrade.

mdbooth:
AWS looks like infra flake?

/test e2e-aws-ovn

mdbooth:
`e2e-azure-ovn` ran successfully, but failed due to slow etcd messages. As this is a known issue on Azure and unrelated to this PR I don't think it's worth re-running it. If it's the only remaining required test in this round of review I would recommend explicitly overriding it to save the CO2.

`e2e-azure-serial` seems to have failed during a control plane rollout? I will investigate this further.

shellyyang1989:
@kyrtapz could you please help review when you have a moment? Thanks!

mdbooth:
@jechen0648 Tested this against a 'public' (`publish: External`) cluster and noted that it was possible to create the full complement of EgressIPs. However, when testing this myself I realised that in fact we have not had this problem on public clusters since #121 merged. The effect of that change was that we didn't add a loadbalancer backend pool to additional IPConfigurations if the rules contain an outbound rule. As they always contain an outbound rule on public clusters, public clusters were not affected.

I take 2 things from this:
* We must test against a private cluster (`publish: Internal`)
* I have increased confidence in this change, as we are already doing this on public clusters

mdbooth:
e2e-azure-serial looks like it's consistently timing out. It has only passed once recently when it came in at 03:55, so just under the limit.

I have pushed https://github.com/openshift/release/pull/65899 to shard it.

PR #179, opened 2025-08-28
Retrieve subnet from vnet subnets.addressPrefixes field

pperiyasamy:
In some Azure clusters where the workers subnet is defined on the vnet using `addressPrefixes` array instead of `addressPrefix`, so retrieve subnet value from vnet's `subnet.addressPrefixes` field when `subnet.addressPrefix` is not set.

### DIFF
--- a/pkg/cloudprovider/azure.go
+++ b/pkg/cloudprovider/azure.go
@@ -3,13 +3,14 @@ package cloudprovider
 import (
 	"context"
 	"fmt"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
-	"k8s.io/utils/ptr"
 	"net"
 	"os"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"k8s.io/utils/ptr"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
@@ -18,7 +19,6 @@ import (
 	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/network/armnetwork/v6"
-	"github.com/Azure/go-autorest/autorest/azure"
 	azureapi "github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/msi-dataplane/pkg/dataplane"
 	v1 "github.com/openshift/api/cloudnetwork/v1"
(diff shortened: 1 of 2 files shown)


pkg/cloudprovider/azure.go
```diff
@@ -577,17 +577,23 @@ func (a *Azure) getAddressPrefixes(networkInterface armnetwork.Interface) ([]*st
 	// FIXME: This might not work for IPv6.
 	if virtualNetwork.Properties != nil && virtualNetwork.Properties.Subnets != nil {
 		for _, vns := range virtualNetwork.Properties.Subnets {
-			if vns.Name != nil && vns.Properties.AddressPrefix != nil &&
-				ptr.Deref(vns.Name, "") == subnetName {
-				return []*string{vns.Properties.AddressPrefix}, nil
+			if vns.Name != nil && ptr.Deref(vns.Name, "") == subnetName {
```
martinkennelly, line 580: so we are trying to network interface addresses+prefixes and if AddressPrefix is set, we chose that. 

What happens if its dualstack cluster and therefore the network interface may have v4 and v6? 
we may only get address + prefix for a single stack. 

Why not favour `AddressPrefixes` over `AddressPrefix` or get the union of the two if the API isnt clearly defined?

martinkennelly:
did you look at how much effort there is to add e2e in origin to highlight this bug? we can then use bot commands to verify this PR using /*-with-pr command to test this and the PR in origin.

we probably need qe ack before merging on azure single and dualstack clusters if its too heavy a lift.

pkg/cloudprovider/azure.go
martinkennelly: how did the reporter end up with `AddressPrefixes` set and not `AddressPrefix` - i assume we have CI for basic sanity testing with a config we support so i am curious how the reporter got the subnet with `AddressPrefixes` set.

pperiyasamy:
> did you look at how much effort there is to add e2e in origin to highlight this bug? we can then use bot commands to verify this PR using /*-with-pr command to test this and the PR in origin.
> 
> we probably need qe ack before merging on azure single and dualstack clusters if its too heavy a lift

AFAIK on existing Azure CI cluster the node subnet is just retrieved from vnet's `addressPrefix` field itself.
so we may need to rely on QE to reproduce and test this fix.

martinkennelly (commented):
/lgtm
/hold for ric

martinkennelly:
> > did you look at how much effort there is to add e2e in origin to highlight this bug? we can then use bot commands to verify this PR using /*-with-pr command to test this and the PR in origin.
> > we probably need qe ack before merging on azure single and dualstack clusters if its too heavy a lift
> 
> AFAIK on existing Azure CI cluster the node subnet is just retrieved from vnet's `addressPrefix` field itself. so we may need to rely on QE to reproduce and test this fix.

And your own manual testing ;) we cannot rely on QE entirely. They are our validation, not primary source of testing.

PR #180, opened 2025-09-26
OCPBUGS-57447: Refrain from adding Egress IP to public LB backend pool

arghosh93:
This PR is to stop adding Egress IP to public load balancer
 backend pool regardless of presence of an OutBoundRule in any
 Azure cluster.
    
This change comes with a consequence of no outbound connectivity
except to the infrastructure subnet even if there is no OutBoundRule.
    
However this is required to tackle following situation:
    
    - If an infra node is being used as an egressNode then health
    check for egress IP also succeeds when it is added to public load
    balancer and LB considers it as a legitimate ingress router backend.
    
    - Limits the number of egress IP which can be created on a cluster
    due to some Azure specific limitation.

this PR also let cobtroller remove any egress IP
added to public load balancer backend pool previously.


arghosh93:
> @arghosh93 Does this PR introduce any limitations on pod egress traffic? From my understanding, if we skip adding the EgressIP to the load balancer backend pools, the egress traffic will be restricted to the infra subnet. Is that correct?

Yes @pperiyasamy , that is correct. The plan is to document this limitation along with a suggestion of using NAT gateway instead of a general purpose public load balancer. I am also gonna notify support team members so that everyone is well aware.

pkg/cloudprovider/azure.go
```diff
@@ -302,6 +301,52 @@ func (a *Azure) GetNodeEgressIPConfiguration(node *corev1.Node, cloudPrivateIPCo
 	return []*NodeEgressIPConfiguration{config}, nil
 }
 
+// The consensus is to not add egress IP to public load balancer
+// backend pool regardless of the presence of an OutBoundRule.
+// During upgrade this function removes any egress IP added to
+// public load balancer backend pool previously.
+func (a *Azure) SyncLBBackend(ip net.IP, node *corev1.Node) error {
```
pperiyasamy: have you tested this `SyncLBBackend`  function with so many EIPs already in place ? because AFAIK Azure APIs are so slow and not sure how it works when you want to sync already existing IPs. 
have you explored sync IPs belong to a node with single API call ? something similar to processing existing items (like [this](https://github.com/ovn-kubernetes/ovn-kubernetes/blob/master/go-controller/pkg/factory/factory.go#L1305-L1322)) before watching CloudPrivateIPConfig objects.

pperiyasamy:
/lgtm
/hold

@kyrtapz is it worth to sync existing CloudPrivateIPConfigs before handling CloudPrivateIPConfig events ? it may optimize number of Azure API calls. Will remove /hold label based on his feedback. Thanks.

pkg/cloudprovider/azure.go
```diff
@@ -357,6 +309,64 @@ func (a *Azure) GetNodeEgressIPConfiguration(node *corev1.Node, cloudPrivateIPCo
...
+	if loadBalanceerBackendPoolModified {
+		networkInterface.Properties.IPConfigurations = ipConfigurations
+		poller, err := a.createOrUpdate(networkInterface)
+		if err != nil {
+			return err
+		}
+		if err = a.waitForCompletion(poller); err != nil {
+			return err
+		}
+		a.lbBackendPoolSynced[cacheKey] = true
```
pperiyasamy: shouldn't this cachedkey cleaned up when ip is deleted ?

pkg/cloudprovider/azure.go
```diff
@@ -357,6 +309,64 @@ func (a *Azure) GetNodeEgressIPConfiguration(node *corev1.Node, cloudPrivateIPCo
...
 }
 
+// The consensus is to not add egress IP to public load balancer
+// backend pool regardless of the presence of an OutBoundRule.
+// During upgrade this function removes any egress IP added to
+// public load balancer backend pool previously.
+func (a *Azure) SyncLBBackend(ip net.IP, node *corev1.Node) error {
+	ipc := ip.String()
+	if a.lbBackendPoolSynced == nil {
+		a.lbBackendPoolSynced = make(map[string]bool)
```
pperiyasamy: it's good to initialize the map while creating Azure CloudProviderClient.

pkg/cloudprovider/azure.go
pperiyasamy: let's wait for @kyrtapz opinion on this.

pkg/cloudprovider/azure.go
```diff
@@ -357,6 +309,64 @@ func (a *Azure) GetNodeEgressIPConfiguration(node *corev1.Node, cloudPrivateIPCo
...
+		}
+	}
+	if loadBalanceerBackendPoolModified {
+		networkInterface.Properties.IPConfigurations = ipConfigurations
+		poller, err := a.createOrUpdate(networkInterface)
+		if err != nil {
+			return err
+		}
+		if err = a.waitForCompletion(poller); err != nil {
+			return err
```
pperiyasamy: same here.

pkg/cloudprovider/azure.go
```diff
@@ -272,11 +215,20 @@ OuterLoop:
...
 	if err != nil {
 		return err
 	}
-	return a.waitForCompletion(poller)
+	if err = a.waitForCompletion(poller); err != nil {
+		return err
+	}
+	// setting lbBackendPoolSynced to true here to make sure that we dont try to
+	// sync LB backend for any new egress IP later
+	cacheKey := fmt.Sprintf("%s|%s", node.Name, ipc)
```
pperiyasamy: this can be moved to `getIPCacheKey` method.

arghosh93: okay

pkg/cloudprovider/azure.go
```diff
@@ -272,11 +216,22 @@ OuterLoop:
 	ipConfigurations = append(ipConfigurations, newIPConfiguration)
 	networkInterface.Properties.IPConfigurations = ipConfigurations
 	// Send the request
+	klog.Warningf("Egress IP %s will have no outbound connectivity except for the infrastructure subnet: "+
```
kyrtapz: As we have discussed in the past this has to have a proper docs in OCP. Is there a card created for it? Are you going to ensure this will be backported to all releases that this change will affect?
This also means that we are a going to significantly change behavior in a Z stream release for the backports so a release note is certainly required too.

kyrtapz (changes requested):
I have some suggestions but I really want to make sure that we have a solid doc plan for this change, what is the version you plan to backport this to?

pkg/controller/cloudprivateipconfig/cloudprivateipconfig_controller.go
```diff
@@ -180,8 +180,14 @@ func (c *CloudPrivateIPConfigController) SyncHandler(key string) error {
 
 	nodeNameToAdd, nodeNameToDel := c.computeOp(cloudPrivateIPConfig)
 	switch {
-	// Dequeue on NOOP, there's nothing to do
 	case nodeNameToAdd == "" && nodeNameToDel == "":
+		node, err := c.nodesLister.Get(cloudPrivateIPConfig.Spec.Node)
+		if err != nil {
+			return err
+		}
+		if err := c.cloudProviderClient.SyncLBBackend(ip, node); err != nil {
```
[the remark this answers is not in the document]
arghosh93: We need cloudProviderClient to run this and that is available with CloudPrivateIPConfigController. Can you please explain if I misunderstood your point?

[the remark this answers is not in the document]
kyrtapz: We are already using cloudProviderClient in the node controller: https://github.com/openshift/cloud-network-config-controller/blob/1456ea2fd35ab7b183cf02c36d09dd5842c2b6e7/pkg/controller/node/node_controller.go#L142

but let's not run the sync on every update as per my previous comment.

pkg/cloudprovider/azure.go
kyrtapz: Thanks, I think that this might be enough.

pkg/cloudprovider/azure.go
```diff
@@ -51,6 +55,10 @@ type Azure struct {
 	nodeMapLock                  sync.Mutex
 	nodeLockMap                  map[string]*sync.Mutex
 	azureWorkloadIdentityEnabled bool
+	cloudPrivateIPConfigLister   cloudnetworklisters.CloudPrivateIPConfigLister
+	nodeLister                   corelisters.NodeLister
+	lbBackendPoolSynced          map[string]bool
```
kyrtapz: What is the point of `lbBackendPoolSynced` and `lbBackendPoolSyncedLock`? Is it a leftover from the previous approach?

pkg/cloudprovider/azure.go
```diff
@@ -357,6 +307,79 @@ func (a *Azure) GetNodeEgressIPConfiguration(node *corev1.Node, cpicIPs sets.Set
...
+			klog.Warningf("source node: %s no longer exists for CloudPrivateIPConfig: %q",
+				cloudPrivateIPConfig.Spec.Node, cloudPrivateIPConfig.Name)
+			continue
+		} else if err != nil {
+			return fmt.Errorf("error getting node %s for CloudPrivateIPConfig %q: %w",
+				cloudPrivateIPConfig.Spec.Node, cloudPrivateIPConfig.Name, err)
+		}
+
+		klog.Infof("Acquiring node lock for modifying load balancer backend pool, node: %s, ip: %s", node.Name, ipc)
+		nodeLock := a.getNodeLock(node.Name)
```
kyrtapz: This is not necessary? If we can remove it the code that follows could be simplified.

PR #182, opened 2025-10-09
OCPBUGS-54805: Handle deleted nodes

kyrtapz:
When a node is removed, the controller would retry indefinitely instead of setting an appropriate error condition. Set error conditions when nodes don't exist rather than retrying indefinitely.


kyrtapz:
@huiran0826 mind helping with the `/verified` label? I know this is just a half of the solution but we need it anyway.

PR #183, opened 2025-10-19
OCPBUGS-60806: Change the capacity struct from int to ptrOfInt

tssurya:
See details on the commit message.

Tested on AWS: Steps: https://issues.redhat.com/browse/OCPBUGS-60806?focusedId=28194587&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-28194587

Before change:
`
    cloud.network.openshift.io/egress-ipconfig: '[{"interface":"eni-00d18740718a0e5d3","ifaddr":{"ipv4":"10.0.0.0/19"},"capacity":{"ipv6":15}}]'
`

After change:
`
    cloud.network.openshift.io/egress-ipconfig: '[{"interface":"eni-00d18740718a0e5d3","ifaddr":{"ipv4":"10.0.0.0/19"},"capacity":{"ipv4":0,"ipv6":15}}]'
`

note that the changes also are backwards compatible with OVN-Kubernetes which uses int and not pointers.
Perhaps a followup should be to also change https://github.com/ovn-kubernetes/ovn-kubernetes/blob/f077fdd127d82bce44a5404a78d4dd88fcf930e5/go-controller/pkg/clustermanager/egressip_controller.go#L1316 into pointers and one more thing to consider is how to not have unlimited capacity on ovn-kubernetes side for cloud since it doesn't make much sense there unlike baremetal. But that is a change in behaviour so that can be another fix.

### DIFF
--- a/pkg/cloudprovider/aws.go
+++ b/pkg/cloudprovider/aws.go
@@ -22,6 +22,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
 	utilnet "k8s.io/utils/net"
+	"k8s.io/utils/ptr"
 )
 
 const (
@@ -218,8 +219,9 @@ func (a *AWS) GetNodeEgressIPConfiguration(node *corev1.Node, cloudPrivateIPConf
 		return nil, err
 	}
 	config.Capacity = capacity{
-		IPv4: capV4,
-		IPv6: capV6,
+		IPv4: ptr.To(capV4),
+		IPv6: ptr.To(capV6),
+		// IP field not used by AWS (uses per-IP-family capacity)
 	}
 	return []*NodeEgressIPConfiguration{config}, nil
 }
--- a/pkg/cloudprovider/azure.go
+++ b/pkg/cloudprovider/azure.go
@@ -3,13 +3,14 @@ package cloudprovider
 import (
 	"context"
 	"fmt"
-	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
-	"k8s.io/utils/ptr"
 	"net"
 	"os"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"k8s.io/utils/ptr"
 
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
 	"github.com/Azure/azure-sdk-for-go/sdk/azcore/arm"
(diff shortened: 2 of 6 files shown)


tssurya:
@kyrtapz PTAL! I tried to add unit tests but not sure why I couldn't find any annotation related tests in CNCC? Maybe I didn't look at the right place?
But LMK what you think of the fix and hopefully I didn't miss any spots wrt changing it to pointer.

@qiowang721 @huiran0826 : could one of you please make sure this fixes the bug 60806 via pre-merge? Also let's make sure test coverage gap gets added to all platforms - aws, gcp, azure, openstack and let's also test upgrades - so reproduce the bug in older version and then upgrade and then make sure the capacity shows up as 0 and then when retrying that it doesn't happen again.

kyrtapz:
> note that the changes also are backwards compatible with OVN-Kubernetes which uses int and not pointers.
Perhaps a followup should be to also change https://github.com/ovn-kubernetes/ovn-kubernetes/blob/f077fdd127d82bce44a5404a78d4dd88fcf930e5/go-controller/pkg/clustermanager/egressip_controller.go#L1316 into pointers 

While I agree that this change shouldn't break ovn-kubernetes I think we should converge the parsing sooner rather than later to avoid future issues, do you mind creating a followup in ovn-k? This shouldn't change the behavior that this change introduces.

>  one more thing to consider is how to not have unlimited capacity on ovn-kubernetes side for cloud since it doesn't make much sense there unlike baremetal. But that is a change in behaviour so that can be another fix.

The behavior is changing with this PR already, once we get it in there is no longer a possibility for a cloud deployment to have an unlimited capacity.
For reference GCP, Azure and OpenStack have predefined static capacity values - meaning that these were never meant to be unlimited anyway.
We read it from the interface config for AWS but looking deeper it is still something that is predefined based on the VM flavor: https://docs.aws.amazon.com/ec2/latest/instancetypes/gp.html#gp_network

So based on the fact that none of the supported CNCC providers allow for unlimited IPs I believe the defaulting in ovn-k is wrong and we should address it as a followup.

qiowang721:
/verified by pre-merge testing

pre-merge tested on fresh installed cluster:
1. create a cluster with PR build
2. Label a node as egress node
3. assign 14 external IPs on node via aws console, then restart CNCC
4. check ipv4 capacity on node, it shows "ipv4":0 as expected
5. create one more EIP, the EIP will not be assigned due to no capacity as expected
6. check cloudprivateipconfigs, there is no entry for the new created EIP
7. scale up another node in the same subnet, and label it with k8s.ovn.org/egress-assignable when it is Ready
8. check the new added EIP will assign to the new added node, cloudprivateipconfig with CloudResponseSuccess

pre-merge tested for upgrade:
1. create a cluster with 4.20 nightly build
2. Label a node as egress node, say node1
3. Configure additional IPs 14 to fill up the capacity
4. Restart cncc, that ipv4 disappeared
5. Create egressIPs 15 egressIPs, and got cloudprivateipconfig with CloudResponseError
6. upgrade cluster to PR build
9. after upgrade, checking the ip capacity for node1, it displays "ipv4":0 as expected

PR #228, opened 2026-06-26
OCPBUGS-87249: on AWS, select associated IPv6 CIDR block for egress IP subnet

jechen0648:
On dualstack IPv4-primary AWS clusters, egressIPv6 addresses are never assigned to egress nodes even though the nodes report non-zero IPv6 capacity in their cloud.network.openshift.io/egress-ipconfig annotation.

The root cause is in getSubnet (pkg/cloudprovider/aws.go). AWS represents an IPv6 CIDR block association on a subnet through Ipv6CidrBlockAssociationSet, and each entry carries a state (associating, associated, disassociated, etc.). A subnet can accumulate multiple entries over its lifetime — for example, after an IPv6 CIDR block is replaced, the old entry stays in the list as disassociated ahead of the new associated entry.

The previous code unconditionally picked Ipv6CidrBlockAssociationSet[0] without checking state, if the first entry is stale (disassociated), one of two things goes wrong:

1) Wrong CIDR returned — the old, now-disassociated IPv6 CIDR is used as the node's IPv6 subnet in the annotation. OVN-Kubernetes stores this and then cannot find a node whose subnet contains the egress IPv6 address (which is from the current, correct CIDR), so it never assigns the EgressIP.
2) No CIDR returned — if the first entry has an empty Ipv6CidrBlock, the condition fails and v6Subnet stays nil. The annotation lacks an IPv6 subnet entirely and OVN-Kubernetes concludes the node has no IPv6 egress capability.
In both cases, no CloudPrivateIPConfig is ever created for the IPv6 EgressIP, so CNCC is never asked to assign it.

Fix
Iterate through all entries in Ipv6CidrBlockAssociationSet and select the first one whose state is associated.

Two minor cleanups are included in the same change:

The empty-subnet guard is tightened from len > 1 to len != 1 so that a DescribeSubnets response returning zero subnets is also caught.
The IPv4 parsing block is cleaned up to avoid shadowing the outer subnet variable (_, v4Net, err instead of _, subnet, err).

Related
A companion fix in [[openshift/ovn-kubernetes](https://github.com/openshift/ovn-kubernetes)](https://github.com/openshift/ovn-kubernetes/pull/3269)  ensures OVN-Kubernetes re-evaluates unassigned EgressIPs when the cloud.network.openshift.io/egress-ipconfig annotation changes, so that any cluster where the annotation was previously incorrect (e.g., missing IPv6 subnet) gets healed automatically once CNCC updates it.

Testing 
Pre-merge testing was performed with https://github.com/openshift/ovn-kubernetes/pull/3269 on dualstack AWS IPv4 primary and dualstack AWS IPv6 primary cluster
Regression testing with https://github.com/openshift/ovn-kubernetes/pull/3269 was performed on dualstack BM cluster


jechen0648:
@coderabbitai review

jechen0648:
@coderabbitai full review

jechen0648:
@arghosh93   I pre-merge tested by fix by building pre-merged image using this PR + https://github.com/openshift/ovn-kubernetes/pull/3269 which is same as u/s PR: https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6648, tests were performed on dualstack AWS for both IPv4 primary and IPv6 primary flavors.

pkg/cloudprovider/aws.go
```diff
@@ -276,23 +276,32 @@ func (a *AWS) getSubnet(networkInterface *ec2.InstanceNetworkInterface) (*net.IP
 	var v4Subnet, v6Subnet *net.IPNet
 	subnet := describeOutput.Subnets[0]
 	if subnet.CidrBlock != nil && *subnet.CidrBlock != "" {
-		_, subnet, err := net.ParseCIDR(*subnet.CidrBlock)
+		_, v4Net, err := net.ParseCIDR(*subnet.CidrBlock)
```
arghosh93, line 279: Why do you have to change the variable name here?

arghosh93 (commented):
Only one NIT cmment and mostly looks god to me.  I will approve based on your answer.

pkg/cloudprovider/aws.go
jechen0648: this subnet variable is used for v4 and v6 below, it's like one variable used for two different ipstack types, it is easy to get confused , that is why I changed the variable name.   But I can revert if you prefer minimal change

arghosh93 (approved):
LGTM

sadasu:
/payload-job-with-prs periodic-ci-openshift-release-main-nightly-5.0-e2e-aws-ovn-installer-dualstack-ipv6-primary-techpreview https://github.com/openshift/ovn-kubernetes/pull/3269

sadasu:
/payload-job-with-prs periodic-ci-openshift-release-main-nightly-5.0-e2e-aws-ovn-installer-dualstack-ipv6-primary-techpreview openshift/ovn-kubernetes#3269

jechen0648:
> @jechen0648 Thanks for this fix. I also see the conformance test `[sig-network][Feature:Router][apigroup:route.openshift.io] The HAProxy router should add replicas beyond the number of empty slots per backend [Suite:openshift/conformance/parallel]` failing on AWS dualstack IPv4 primary and IPv6 primary jobs. I opened https://redhat.atlassian.net/browse/OCPBUGS-98795 for that issue. Would this fix also take care of that issue?

I do not think so, `[sig-network][Feature:Router][apigroup:route.openshift.io] The HAProxy router should add replicas beyond the number of empty slots per backend [Suite:openshift/conformance/parallel]`  is a test from network edge team, I do not cover this area.  My fix is specifically for egressIP feature that is a core networking feature

pkg/cloudprovider/aws.go
pperiyasamy, line 1: so previously does it took 'disassociated' subnet entry at [0] which is different from host subnet, what was that subnet value ? just curious.

pkg/cloudprovider/aws.go
```diff
@@ -276,23 +276,32 @@ func (a *AWS) getSubnet(networkInterface *ec2.InstanceNetworkInterface) (*net.IP
...
+	// "disassociated" one and a current "associated" one). On dualstack AWS clusters
+	// after subnet CIDR changes, picking the wrong (stale) CIDR causes egress IP
+	// assignment to fail because OVN-Kubernetes selects IPs from the wrong range.
+	// Always select the first association in "associated" state.
+	for _, assoc := range subnet.Ipv6CidrBlockAssociationSet {
+		if assoc.Ipv6CidrBlock == nil || *assoc.Ipv6CidrBlock == "" {
+			continue
+		}
+		if assoc.Ipv6CidrBlockState == nil || assoc.Ipv6CidrBlockState.State == nil ||
+			*assoc.Ipv6CidrBlockState.State != ec2.SubnetCidrBlockStateCodeAssociated {
```
pperiyasamy, line 296: is there a possibility of more than one associated subnet ? hope you have checked it.

pperiyasamy:
> The root cause is in getSubnet (pkg/cloudprovider/aws.go). AWS represents an IPv6 CIDR block association on a subnet through Ipv6CidrBlockAssociationSet, and each entry carries a state (associating, associated, disassociated, etc.). A subnet can accumulate multiple entries over its lifetime — for example, after an IPv6 CIDR block is replaced, the old entry stays in the list as disassociated ahead of the new associated entry.

you mentions a subnet can accumulate multiple entries over lifetime, do you mean lifetime of the node ? I assume associated ipv6 cidr never changes. right ? because cncc needs to populate  `cloud.network.openshift.io/egress-ipconfig`  annotation at the time of first node add event and EIPs can be assigned to it or is associated ipv6 cidr itself subject to change ?

jechen0648:
> > The root cause is in getSubnet (pkg/cloudprovider/aws.go). AWS represents an IPv6 CIDR block association on a subnet through Ipv6CidrBlockAssociationSet, and each entry carries a state (associating, associated, disassociated, etc.). A subnet can accumulate multiple entries over its lifetime — for example, after an IPv6 CIDR block is replaced, the old entry stays in the list as disassociated ahead of the new associated entry.
> 
> you mentions a subnet can accumulate multiple entries over lifetime, do you mean lifetime of the node ? I assume associated ipv6 cidr never changes. right ? because cncc needs to populate `cloud.network.openshift.io/egress-ipconfig` annotation at the time of first node add event and EIPs can be assigned to it or is associated ipv6 cidr itself subject to change ?

It's the lifetime of the subnet, not the node.  A subnet is a range of IP addresses defined in VPC. It exists at the network/infrastructure level. You create it once when setting up the cluster and it keeps existing regardless of what happens to individual machines. A node (EC2 instance) is a virtual machine that runs inside a subnet. Nodes come and go — they get terminated and replaced during upgrades, scaling events, or failures. subnet persists independently, any historical changes to it (like a past IPv6 CIDR replacement) are still visible in the AWS API even much time later, even to brand new nodes joining that subnet.  The stale disassociated entry in Ipv6CidrBlockAssociationSet could be from a past change to the subnet,  it had nothing to do with the specific node CNCC was looking at.

pkg/cloudprovider/aws.go
jechen0648: Yes, technically AWS allows multiple IPv6 CIDR blocks in associated state simultaneously (e.g., when additional IPv6 prefixes are added to a subnet). The current code takes the first associated entry, silently ignoring any others.  This means EgressIPs can only be assigned from the first CIDR. This is consistent with how we handle the node's subnet today.   If multiple associated entries exist,  should I add a klog.Warningf when a second associated entry is found to make this situation visible in logs?   Is that what you expect?

pkg/cloudprovider/aws.go
jechen0648: I created a draft [PR](https://github.com/openshift/cloud-network-config-controller/pull/261) trying to print out of the first IPv6 CIDR, I ran egressIP test on dualstack AWS a few times, I have not been able to reproduce the first entry as disassociated one.    OCPBUGS-87249 is not always reproducible.

pkg/cloudprovider/aws.go
pperiyasamy: is first associated CIDR entry for node subnet? wanted to confirm that, otherwise we may end up EIP not getting assigned to the node.

sadasu:
/payload-job-with-prs periodic-ci-openshift-release-main-nightly-5.0-e2e-aws-ovn-installer-dualstack-ipv6-primary-techpreview https://github.com/openshift/ovn-kubernetes/pull/3269

pkg/cloudprovider/aws.go
jechen0648: yes, it is from node’s subnet, the sequence is
getInstance() gets EC instance
then
getNetworkInterfaces() get EC's networkInterface
then  we get v4 or v6 CIDR from line 204:
   v4Subnet, v6Subnet, err := a.getSubnet(networkInterface)

jechen0648: BTW, regarding whether there would be multiple v6 CIDR blocks, I got more accurate answer from @sadasu:  within a VPC in which the cluster is installed. there could be 3 Availability zones (AZs). Within each AZ, there needs to be 1 public and 1 private subnet. So, with this calculation, we should have 6 IPv6 CIDRs within an install.
So, when we consider Egress IPs on worker nodes, for worker would be associated with 1 AZ. And within that there would be 1 public subnet. So, for a worker node used for egress OP should have only 1 active IPv6 CIDR.

sadasu:
/payload-job-with-prs periodic-ci-openshift-release-main-nightly-5.0-e2e-aws-ovn-installer-dualstack-ipv6-primary-techpreview openshift/ovn-kubernetes#3269

pkg/cloudprovider/aws.go
jechen0648: @pperiyasamy Regarding to your question of "Is there a chance private subnet present at 0th index ?", I am not expert of AWS, this is what I got after asking for AI:

each node will be in 1 AZ, there is 1 public and 1 private subnet for each AZ.  On a standard (non-private-only) OpenShift AWS cluster:
1. worker and control plane nodes are usually placed in private subnets
2. Public subnets are mainly for NAT gateway, load balancers and similar edge infrastructure.

For EgressIP, relevant IPv6 CIDR is normally the private subnet in that worker's AZ.
Each egress-capable worker exposes 1 active associated IPv6 CIDR on its node annotation - the one for the subnet its ENI is actually in.

CNCC does not pick "public Vs private", it reads whichever subnet the node's primary ENI is attached to.

OVNK choose egressIP from that annotated CIDR, so:
if worker is in AZ's private subnet, egressIP is from the private subnet/64
if worker is in AZ's public subnet, egressIP is from the public subnet/64

"private subnet" is an AWS routing concept.  egressIP address is additional IP on the instance ENI from that node's subnet CIDR.    CNCC assigns it via AssignIPv6Addresses regardless of public/private subnet classification.

@sadasu can you help to see if the above explanation makes sense to you?  Thanks!

pperiyasamy: ok @jechen0648 , so i understand node subnet is from private subnet, now cncc chooses first assoicated from  `subnet.Ipv6CidrBlockAssociationSet` for the egressip-config annotation, so is that always private subnet set as first item ?

jechen0648: No — Ipv6CidrBlockAssociationSet is per-subnet, not a mix of public and private subnets.

If the worker is in the private subnet (for most normal OpenShift clusters), CNCC describes that subnet via the node ENI’s SubnetId, and the annotation gets that private subnet’s IPv6 /64. We pick the first associated IPv6 CIDR on that subnet, skipping any disassociated (stale) entries.

If the worker were placed in a public subnet, the same logic applies and the annotation would get that public subnet’s IPv6 /64 instead.

Ipv6CidrBlockAssociationSet on a given subnet only contains IPv6 CIDR blocks for that subnet — it does not mix public and private subnet CIDRs.

sadasu: @jechen0648 +1 . Above approach makes sense. 
Is `cloud.network.openshift.io/egress-ipconfig` the annotation are we referring to here?

sadasu: > is there a possibility of more than one associated subnet ? hope you have checked it.

1 NIC associated with the primary subnet on the worker node is associated with OVNK. I think we are just interested in the subnet associated with this NIC and finding its IPv6 subnet that is still associated.

pkg/cloudprovider/aws.go
```diff
@@ -276,23 +276,32 @@ func (a *AWS) getSubnet(networkInterface *ec2.InstanceNetworkInterface) (*net.IP
...
+			continue
+		}
+		_, v6Net, err := net.ParseCIDR(*assoc.Ipv6CidrBlock)
 		if err != nil {
 			return nil, nil, fmt.Errorf("error: unable to parse IPv6 subnet, err: %v", err)
 		}
-		v6Subnet = subnet
+		v6Subnet = v6Net
+		break
```
sadasu, line 304: If this method is called only for dualstack installations, we should warn when we don't find any IPv6 subnet that is associated. For DualStack IPv6 primary, that would be an error and IPv4Primary that would be a warning. Is that captured anywhere?

sadasu:
I agree that the bug is hard to reproduce. The results from https://pr-payload-tests.ci.openshift.org/runs/ci/aaa99e80-90d6-11f1-9bba-3026d0a99078-0 show there are no regressions from this.

pkg/cloudprovider/aws.go
jechen0648: > @jechen0648 +1 . Above approach makes sense. Is `cloud.network.openshift.io/egress-ipconfig` the annotation are we referring to here?

yes

pkg/cloudprovider/aws.go
jechen0648: getSubnet() will be called for any IP stack type AWS, if Ipv6CidrBlockAssociationSet is not empty, it just processes and finds the first associated IPv6 CidrBlock

You're right that on dual-stack clusters, a missing associated IPv6 subnet may warrant a warning or error depending on Infrastructure.status.platformStatus.aws.ipFamily (DualStackIPv6Primary vs DualStackIPv4Primary). That logic is not implemented today — CNCC does not read cluster IP family config. 

I think that's out of scope for this PR, which only fixes selecting the correct associated IPv6 CIDR when multiple associations exist. How about to open a follow-up for IP-stack-aware validation?

sadasu: Understood. Thanks!

sadasu:
@pperiyasamy @kyrtapz, @tssurya This seems ready for approval.

pkg/cloudprovider/aws.go
pperiyasamy: thanks @jechen0648 and @sadasu !

PR #230, opened 2026-07-15
OSASINFRA-4400: make openstack allowed address pairs configurable

danchild:
Currently, the **maxmium allowed address pairs** per port configuration
is hardcoded to OpenStack's default of 10. The new
`platform-os-max-allowed-address-pairs` flag allows
users to customize the configuration, mirroring the
setting in their installation's **neutron.conf**.

The patch allows any custom configuration greater than
zero and disallows anything less than or equal to zero.
Finally, the default of 10 is still used as a fallback
if the user doesn't configure `platform-os-max-allowed-address-pairs`


IlanZuckerman:
The Jira ticket says

> When Neutron rejects an assignment with AllowedAddressPairExhausted (HTTP 400), surface a specific actionable error

Are we implementing that in this PR?

Also, pls consider adjusting the PR description. It says:

> ... and disallows anything less than or equal to zero.

I think we are NOT dissalowing here, but falling back to the defaults.

winiciusallan:
/lgtm 

I'd probably return an error (or at least log a message) when the default passed by the user is less than zero. One may have the impression that the value was applied correctly, but it is not acceptable by the API.

danchild:
> /lgtm
> 
> I'd probably return an error (or at least log a message) when the default passed by the user is less than zero. One may have the impression that the value was applied correctly, but it is not acceptable by the API.

Good thinking here. Because the controller is a consumer of the ConfigMap config passed upstream from the `cluster-network-operator` here [openshift/cluster-network-operator/pull/3058](https://github.com/openshift/cluster-network-operator/pull/3058), negative values won't pass through. I figured that creating a check here would be duplicative.

danchild:
> The Jira ticket says
> 
> > When Neutron rejects an assignment with AllowedAddressPairExhausted (HTTP 400), surface a specific actionable error
> 
> Are we implementing that in this PR?
> 
> Also, pls consider adjusting the PR description. It says:
> 
> > ... and disallows anything less than or equal to zero.
> 
> I think we are NOT dissalowing here, but falling back to the defaults.

Good catch on the JIRA ticket

I addressed a similar guard here https://github.com/openshift/cloud-network-config-controller/pull/230/changes#diff-3e4b529b8234de46127ef3820b6f3514f1dfc1cc694749b26edc2372eece58b3R610. I figured was a simpler solution to the same problem which utilizes existing code.

cmd/cloud-network-config-controller/main.go
```diff
@@ -264,6 +264,7 @@ func init() {
 	flag.StringVar(&platformCfg.AzureEnvironment, "platform-azure-environment", "AzurePublicCloud", "The Azure environment name, used to select API endpoints")
 	flag.StringVar(&platformCfg.AWSCAOverride, "platform-aws-ca-override", "", "Path to a separate CA bundle to use when connecting to the AWS API")
 	flag.StringVar(&kubeConfig, "kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
+	flag.IntVar(&platformCfg.OSMaxAllowedAddressPairs, "platform-os-max-allowed-address-pairs", 0, "Maximum number of allowed address pairs per OpenStack port. Overrides the Neutron default of 10.")
```
stephenfin: Any reason for `OS` instead of `OpenStack`?

pkg/cloudprovider/openstack.go
```diff
@@ -51,7 +51,7 @@ const (
 	// to the default value of 10, always; and we might want to document that OSP environments must set
 	// max_allowed_address_pairs >= 10. For more details, see:
 	// https://github.com/openstack/neutron/blob/800f863ccc502b334cb2dd79ec54066440e43e27/neutron/conf/extensions/allowedaddresspairs.py#L21
-	openstackMaxCapacity = 10
+	defaultOpenstackMaxCapacity = 10
```
stephenfin: nit: 
```suggestion
	defaultOpenStackMaxCapacity = 10
```

winiciusallan (commented):
/lgtm

tssurya:
/hold

as discussed https://redhat-internal.slack.com/archives/CDCP2LA9L/p1786360947298439?thread_ts=1784294783.765949&cid=CDCP2LA9L

tssurya:
/hold cancel

https://redhat-internal.slack.com/archives/CDCP2LA9L/p1790070719183859?thread_ts=1789566928.504719&cid=CDCP2LA9L

since the PRs are small enough it shouldn't be much effort, leaving it to the bot assigned reviewers

arkadeepsen:
@coderabbitai review

cmd/cloud-network-config-controller/main.go
```diff
@@ -264,6 +264,7 @@ func init() {
 	flag.StringVar(&platformCfg.AzureEnvironment, "platform-azure-environment", "AzurePublicCloud", "The Azure environment name, used to select API endpoints")
 	flag.StringVar(&platformCfg.AWSCAOverride, "platform-aws-ca-override", "", "Path to a separate CA bundle to use when connecting to the AWS API")
 	flag.StringVar(&kubeConfig, "kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
+	flag.IntVar(&platformCfg.OpenStackMaxAllowedAddressPairs, "platform-os-max-allowed-address-pairs", 0, "Maximum number of allowed address pairs per OpenStack port. Overrides the Neutron default of 10.")
```
arkadeepsen:
```suggestion
	flag.IntVar(&platformCfg.OpenStackMaxAllowedAddressPairs, "platform-os-max-allowed-address-pairs", 10, "Maximum number of allowed address pairs per OpenStack port. Overrides the Neutron default of 10.")
```
Do you think we can add the default value as 10 here itself?

arkadeepsen (commented):
Apart from the following comment, rest looks good.

cmd/cloud-network-config-controller/main.go
danchild: Yes of course. Note that doing so would make the following redundant:

`defaultOpenStackMaxCapacity` (because we're setting the default in the flag)

and ...

```
if o.cfg.OpenStackMaxAllowedAddressPairs > 0 {
		openstackMaxCapacity = o.cfg.OpenStackMaxAllowedAddressPairs
} else {
		openstackMaxCapacity = defaultOpenStackMaxCapacity
}
```  

because we know that `o.cfg.OpenStackMaxAllowedAddressPairs` will always be greater than 0. In the related PR https://github.com/openshift/cluster-network-operator/pull/3058, we ensure that 0 or negative values will never be passed.

What do you think?

arkadeepsen: If we specifically set `o.cfg.OpenStackMaxAllowedAddressPairs` to 0 then that value will be used, rather than defaulting to 10. In the case of where this flag is not set, then default value of 10 will be set. So the code snippet that you mentioned above will still be needed.

Additionally, it's good to have local validations even if CNO is the only place from where it is being set. This keeps the behavior consistent even if some changes to CNO. If a change in behavior is needed, it has to be specifically changed in CNCC as well.

